PR #11, opened 2026-07-01
feat(ui): wire per-adapter model switching into the composer

lucas77778:
Replace the composer's decorative model-cycling button with a real picker driven by AGENT_MODEL_OPTIONS, keyed by AgentKind — only rendered for adapters with a verified live switch (currently just claude-code; opencode/pi have no entry and stay hidden until verified). Threads agentKind (already available one level up) and onModelChange down through WorkbenchConversationSurface -> WorkbenchFrame / desktop shell -> Workbench, backed by the new client-core/sdk setModel() wrappers on top of fix/claude-code-live-model-switch's `set-model` AgentInput.


lucas77778:
## macOS build (\`Build (mac)\`) — EMFILE diagnosis, unresolved

While fixing the two other CI failures on this PR (missing `icon-dock.png` asset — fixed in 972b8a7, now green; TypeScript/lint now green), the macOS packaging job is still failing with:

```
⨯ error parsing dependencies tree  error=Node module collector process exited with code 232:
{ "error": { "code": "EMFILE" } }
```

This is **not caused by this PR's diff** — the identical failure occurs on unrelated branches too (e.g. `feat/terminal-pty-sidecar` run [28511202191](https://github.com/arcboxlabs/linkcode/actions/runs/28511202191)), so it looks like a pre-existing, repo-wide issue with the `Build (mac)` job that's been failing consistently for at least the last day across multiple PRs.

**What I tried:**
- Added `ulimit -n 10240` before the `electron-builder` step (the value electron-builder's own docs recommend for this exact error) — commit 1b78fd0. Still failed, identically.
- Raised it to `ulimit -n 65536` — commit 08373dc. Still failed, identically, and in almost exactly the same (short, ~1.3s) time regardless of the limit used.
- Reproduced the *exact* CI command locally (`pnpm exec electron-builder --mac --publish never`, matching `mac.target.arch: [universal]` in `electron-builder.yml`) on a real macOS machine — the node-module-collector step **succeeds** every time locally, at any ulimit including the default. Locally the build only fails later, on an unrelated `actool`/Xcode version check (a local-machine-only issue, not present on CI's runner which presumably has the right Xcode).

**Conclusion:** the ulimit fix I applied is a legitimate, low-risk improvement (kept in place), but it isn't the actual bottleneck here — the CI runner's environment seems to differ from a normal macOS machine in some way that isn't just "too low a soft ulimit" (possibly a lower effective hard limit specific to the hosted runner image, or something specific to a fresh CI checkout's `pnpm install` vs. a long-lived local store). I wasn't able to pin down the exact mechanism without direct access to the runner.

**Suggested next steps (need a human call, didn't want to guess further):**
- Try pinning an older `electron-builder` version (this is `26.15.3` — worth checking its changelog/issues for pnpm + node-module-collector EMFILE regressions).
- Or try building the two architectures separately (`--x64` / `--arm64`) instead of `arch: [universal]`, to see if a single-arch build avoids whatever is doubling up file handles.
- Or escalate to GitHub Actions / Blacksmith support with the exact repro above if this is runner-image-specific.

Everything else on this PR is green (TypeScript, lint, All Green, Greptile). This is the one remaining blocker.

PR #52, opened 2026-07-08
feat(agent): read the claude-code model catalog from the SDK (wire v15)

AprilNEA:
Closes the claude-code model slice of CODE-104.

## What

Replaces the hand-curated static model table (`AGENT_MODEL_OPTIONS` in `packages/ui/src/shell/agent-models.ts`, now deleted) with the CLI's own model catalog, read live off the claude-agent-sdk and served to every client over a new pull wire resource.

- **schema** — new `agent-model.list → agent-model.listed` resource (`AgentModels = Partial<Record<AgentKind, {id, label}[]>>`), mirroring `agent-runtime.list`; **`WIRE_PROTOCOL_VERSION` 14 → 15** (restart daemon + clients together).
- **agent-adapter** — `AgentAdapter.listModels(config?)`, base default `[]`. claude-code reads `Query#supportedModels()` off a transient never-prompted streaming `Query` (the `initialize` control request fires at Query construction, so no prompt is needed), closed as soon as the catalog arrives; 30s timeout since plain `query()` applies none; same `apiKey`-env injection as sessions.
- **engine / client-core / sdk** — engine probes lazily per kind with the stored provider config, caches per boot, and evicts a rejected probe so a later request retries (e.g. after installing/logging into the CLI); a failed kind is absent from the map instead of failing the reply. Full request plumbing mirrors `agent-runtime`.
- **workbench / ui / desktop** — `useAgentModels()` fetched at Workbench mount (pre-warms the probe before a menu ever opens), threaded as props into both composers (conversation + new-session); dev mock host serves a fixture catalog. The effort table stays static (CODE-104's codex/per-model-efforts tasks remain).

## Why dynamic beats the curated table

Verified live, the CLI registry is **entitlement-aware and alias-based**: an authed account gets `default / opus[1m] / claude-fable-5[1m] / sonnet / haiku` (display names from the vendor), while an unauthenticated environment's catalog lacks fable. The static table could neither track new releases nor reflect per-account access, and its pinned legacy ids (opus-4-7/4-6, sonnet-4-6) don't exist in the CLI's own /model menu at all.

Accepted trade-off (recorded in `packages/agent-adapter/AGENTS.md` + CODE-104): the catalog mirrors the CLI's menu rather than per-entry live-switch verification, so vendor-side quirks (e.g. the documented opus-4-1 silent substitution) are vendor behavior, not filtered locally.

## Verification

- `pnpm check:ci` and full `pnpm test` green (396 tests; 8 new — 4 engine wire/caching/retry/config, 4 adapter mapping/env/timeout/close).
- Live adapter probe against the real CLI returned the real catalog in ~1–4s.
- Live wire round-trip through an isolated daemon (`HOME=<tmp> LINKCODE_PORT=19777`) over a real socket.io connection returned the catalog end-to-end; `initialize` answers without credentials, so the probe needs no auth.


AprilNEA:
Closing unmerged by explicit decision: models change rarely, so the hand-curated static table is the better trade — a per-boot CLI spawn (and the extra wire resource + probe machinery) isn't justified just to keep a short list fresh. The static `AGENT_MODEL_OPTIONS` in `packages/ui/src/shell/agent-models.ts` stays the deliberate source of truth.

Keeping this PR as the reference implementation if the trade ever flips (it was fully verified: live catalog probe + wire round-trip). Durable findings recorded on CODE-104: the CLI registry serves alias ids (`opus[1m]`, `default`) rather than pinned snapshots, is entitlement-aware per account, and `initialize` answers without auth at Query construction.

PR #57, opened 2026-07-08
feat(desktop,workbench,engine): OS notifications for background sessions

AprilNEA:
Closes CODE-116.

## What

End-to-end OS notifications: a background session finishing a turn, awaiting approval, or erroring now raises a system notification on desktop (main-process `Notification`) and webview (Web Notification API); clicking it focuses the window and jumps to the session.

- **schema/engine (wire v15→16)**: new `session.notification` broadcast, classified daemon-side — `stop` → turn-completed (`status:'idle'` also fires at session start, so it can't be the trigger), `permission-request` → awaiting-approval (no adapter ever emits an `awaiting-input` status), `error` → error. Carries its own display fields (kind/cwd/title) since a client may not have the session in its list snapshot. Stays a broadcast even once per-connection subscription modes exist (CODE-72).
- **client-core/sdk**: `subscribeSessionNotification`, mirroring the `script.status` broadcast pattern.
- **workbench**: `shouldPresent` policy (suppress when window focused and session active; never notify user-initiated cancels; per-reason prefs) + `zodPersist` prefs store (`linkcode.workbench.notifications:v1`) + i18n. Presenters are app-injected via a new `Workbench` prop, keeping the decision layer shared.
- **desktop/ipc**: SystemBridge `notifications` capability. `notify()` carries only title/body/`clickToken` — an opaque echo the IPC layer doesn't interpret, so the system plane stays business-data-free; main shows the notification, and a click restores/focuses the window then pushes the token back for selection.
- **webview**: Web Notification presenter (permission request/denied UX in settings).
- **settings**: shared `NotificationsSettingsPanel` wired into the desktop overlay tab and a webview route.
- **fix**: the new-session draft moved from `useWorkbenchSessions` local state into the selection store — an outside `setSelectedId` (notification click, palette) now atomically exits an open draft instead of staying masked by it.

## Verification

- Engine classification integration tests, client fan-out test, `shouldPresent` unit tests; `pnpm check:ci` + full vitest suite green.
- First committed E2E: `pnpm -F @linkcode/desktop e2e:notifications` (isolated daemon + built app, main-side `Notification.prototype.show` stub, `emit('click')` click-through assertion; requires `--use-mock-keychain`, now baked in — docs updated in DEVELOPMENT.md).

## Reviewer note

Wire protocol bumped to **v19** (rebased over master's v18): daemon and every client must be rebuilt/restarted together; a v15 peer connects but silently drops every frame (Invariant 1).


AprilNEA:
Rebased onto master (twice — CODE-118/119/onboarding wire bumps, then CODE-107 packaging) and addressed the review findings.

**Conflicts** — wire is now **v19** (master reached v18; this PR adds the `session.notification` variant on top). Folded master's new `question-request` ask type into the notification classifier: `question-request` now raises an `awaiting-approval` notification alongside `permission-request` (test added).

**Codex/Greptile — click-through selects the wrong thread when the session isn't in the local list** (`notifications.ts`): fixed in the data layer so it covers every selection source (notification, palette, future deep-links), not just this one call site. `useWorkbenchSessions` no longer falls back to a different thread when an explicit `selectedId` is absent from the loaded list — it holds and refreshes `listSessions` once (deduped per id), so a click-through to a session another client created resolves to the right conversation instead of a stale one.

**Codex — webview drops notifications while on `/settings`** (`workbench-route.tsx`): the subscriber moved out of the routed `Workbench` into a headless `SessionNotifier` mounted in each app's persistent layer (desktop's permanent shell, webview's `RootLayout` below the connection gate), so navigating away no longer unmounts it. Suppression now keys on the explicitly-selected session (selection store) instead of the routed workbench's resolved active id, which is what lets it live in the persistent layer.

`pnpm check:ci` + full vitest (492 passing) green after both merges.

PR #66, opened 2026-07-08
fix(desktop): Windows/Linux window controls + taskbar icon identity

AprilNEA:
Two Windows-desktop chrome fixes reported on a Windows (Parallels) build.

## Problem 2 — no minimize/maximize/close buttons on Windows/Linux
`titleBarStyle: 'hidden'` keeps macOS's native traffic lights but strips the OS caption buttons on Windows/Linux, and the renderer never drew replacements. The window IPC bridge (`systemBridge.window.minimize/toggleMaximize/close/isMaximized/onMaximizedChange`, plus the `WINDOW_MAXIMIZED_CHANGED_CHANNEL` push) was already wired end-to-end with **zero consumers** — the original design clearly intended renderer-drawn buttons.

- New `window-controls.tsx`: minimize / maximize-restore / close as ghost `ShellIconButton`s; close hovers red; the restore icon toggles off the existing `onMaximizedChange` push.
- Rendered in the chrome right rail (left of the panel toggles, counted in the rail-inset measurement so it never underlaps the title).
- `showWindowControls = platform known && !== 'darwin'` — gated on a resolved platform so macOS never flashes them before `app.platform()` returns. Both the main shell and the Settings overlay get them.
- `packages/ipc` `bridge.ts`: `onMaximizedChange` promoted from optional to required, matching its sibling subscriptions and its sole implementation.

## Problem 1 — blank taskbar icon on Windows
Added the missing `app.setAppUserModelId(APP_ID)` (win32 only) — Windows keys the taskbar icon, pinning, and notification identity off it. Packaged builds still get their icon from electron-builder (`icon.png` → ICO); a dev-shell blank in a VM may also be a Parallels icon-cache artifact.

## Verify on Windows (couldn't run Windows locally)
- [ ] ⊟ ▢ ✕ appear top-right; minimize/close work; maximize toggles to the restore glyph and back
- [ ] close hovers red; drag-region moves the window; double-click maximizes/restores
- [ ] macOS shows **no** right-side controls (still left traffic lights), no flash on launch
- [ ] taskbar icon no longer blank (if a **packaged** build still blanks, ship a dedicated multi-size `.ico`)

Verified: typecheck / eslint (0 errors) / biome / 486 tests all pass.

### DIFF
--- a/apps/desktop/src/main/constants.ts
+++ b/apps/desktop/src/main/constants.ts
@@ -13,5 +13,15 @@ export const IS_DEV_SHELL = import.meta.env.MODE !== 'production' || !app.isPack
 
 export const APP_NAME = IS_DEV_SHELL ? 'LinkCode Dev' : 'LinkCode';
 
+/**
+ * Windows AppUserModelID — mirrors the electron-builder `appId`. Windows keys the taskbar icon,
+ * pinning, and notification identity off this; without setting it at runtime the taskbar shows a
+ * blank/default icon. A dev shell gets a distinct id so it neither steals the installed release's
+ * taskbar slot nor its notifications (same isolation rationale as `APP_NAME`).
+ */
+export const APP_ID = IS_DEV_SHELL
+  ? 'com.arcboxlabs.linkcode.desktop.dev'
+  : 'com.arcboxlabs.linkcode.desktop';
+
 /** `~/LinkCode` holds user workspaces — shared between dev and release shells on purpose. */
 export const DEFAULT_WORKSPACES_DIRNAME = 'LinkCode';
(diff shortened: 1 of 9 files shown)


AprilNEA:
Addressed the Codex P2 (window controls only mounted inside `DesktopShell`, below the connection gate) in `016c3a2`.

The window controls are now a **persistent layer** (`DesktopWindowControls`) mounted at the `DesktopApp` root — above both the connection gate and the settings overlay (`z-[60]`, > settings' `z-50`). They stay present on Windows/Linux through every state: `ConnectionSkeleton`, `ConnectionState` (bad URL / daemon down), the shell, and settings — and never shift position as the app moves between them.

- Removed the controls from the chrome's right rail; the chrome now reserves an equal-width inset (`--lc-chrome-window-controls-inset`, mirroring the macOS `NativeTrafficLightInset`) so the panel toggles never sit under the floating controls.
- Still gated on a resolved non-macOS platform, so macOS keeps its native traffic lights and never flashes.

Verified: desktop typecheck / eslint (0 errors) / biome / tests (the 2 unrelated daemon-port + npm-registry failures are pre-existing environmental flakiness; they pass in isolation).

PR #71, opened 2026-07-09
feat: Account system — multi-account switching, provider conversion, AI gateway, Providers settings UI

AprilNEA:
## Account system — multi-account switching + provider conversion + AI gateway + settings UI

Implements the CODE-127 tree end-to-end: keep **multiple credentials per coding agent** in a global pool and switch between them, route a credential from one wire protocol into a different agent (e.g. an OpenAI/OpenRouter key backing Claude Code), gateway presets, and a full **Providers × Agents settings surface** in both desktop and webview. Design docs: [architecture](https://claude.ai/code/artifact/b873690d-e905-48ec-9b0e-41c5f23494f5) · [settings UI (v2.1)](https://claude.ai/code/artifact/00240a81-e9c9-4bb5-8742-82eb81aa3d94)

Closes CODE-128 · related CODE-127 / CODE-129 / CODE-130 / CODE-131 / CODE-132 / CODE-133.

### Architectural key
LinkCode owns the spawn and injects credentials **per session** (`applyProviderDefaults` → `opts.config` → each adapter's env/options), so account switching is per-session and race-free — **no cc-switch-style rewriting of the CLI's own global config files**. Two sessions can run two accounts at once.

### What's here (21 commits, layer by layer)
**Data plane & translation**
- **schema** — global `Account` pool (+ `service` catalog key) + `providers[kind].activeAccountId`; `config.get/set` carry `accounts`; daemon persists them (`0600`, field-by-field validation); `AgentAuthStatus.email`.
- **engine** — account-resolving `applyProviderDefaults` (`accountId → activeAccountId → legacy apiKey`); cross-protocol accounts routed through a translation sidecar.
- **agent-adapter** — per-agent injection of the resolved credential + base URL (claude env-replace incl. the `ANTHROPIC_API_KEY=""` gateway gotcha; codex env-merge; opencode `options.baseURL`; pi `registerProvider`); the login probe now surfaces the signed-in **email**.
- **daemon** — spawns `arcboxlabs/aigateway` as a loopback sidecar for offline Anthropic⇄OpenAI translation; the binary ships as a **managed asset** (`tool:aigateway`, pinned v0.5.0-rc2, per-platform SRI) resolved env → asset store.
- **sdk / client-core** — `getAccounts` / `setAccounts` over the config channel (additive).

**Settings UI (shared workbench containers, mounted by desktop + webview)**
[...]

PR #76, opened 2026-07-10
fix(agent-adapter): handle claude-code compaction so history isn't lost on auto-compact

lucas77778:
Closes CODE-141.

## What

After a claude-code context compaction (auto or manual `/compact`), LinkCode now shows a "Context compacted" marker in the timeline (with pre/post token counts and the expandable swapped-in summary), and the pre-compaction turns remain visible — both live and when the session history is re-read.

## Findings (verified live against SDK 0.3.179)

- **The open question is answered: `session_id` does NOT change across a compaction in streaming-input mode.** The session-rebind suspect in the issue is disproven — no continuation guard is needed.
- The real cause of the "summary handed off to a new session, history gone" symptom is on the **history-read path**: `getSessionMessages`'s chain reconstruction starts at the newest compaction summary (`parentUuid: null`; `logicalParentUuid` is ignored), so every pre-compaction row silently vanishes from the read — and the swapped-in summary row replayed as a giant fake *user prompt*.
- On the live stream the summary arrives as an `isReplay`+`isSynthetic` user frame whose uuid equals the boundary's `anchor_uuid`, so it was already being silently dropped by the replay guard (that's why no marker ever appeared live).
- The SDK read API also strips the boundary row's `subtype`/`compactMetadata` and the summary row's `isCompactSummary` flag, so the raw transcript is the only carrier of compaction facts on re-read.

## How

- **schema**: new `compaction` `AgentEvent` variant (`compactionId`, `trigger`, `preTokens`, `postTokens`, `summary`); consumers merge partial emits by `compactionId`. `WIRE_PROTOCOL_VERSION` 23 → 24 (rebased over master, which had moved to 23).
- **agent-adapter (claude-code)**:
  - live: `compact_boundary` → emit the marker immediately; catch the summary frame before the replay guard (matched by anchor uuid, `isSynthetic` fallback) and re-emit with `summary`.
  - history: `readClaudeCompactionSupplement` reads the raw transcript once per `readHistory` — an index that turns summary rows into markers instead of fake user prompts, plus the pre-compaction rows the SDK drops, prepended on the first page and deduped by uuid against the preserved segment.
- **client-core / ui / i18n**: `compaction` conversation item merged by id; `CompactionMarker` divider (token counts + collapsible summary) in the timeline.

## Verification

- Drove the real adapter through a live session with `/compact` (real SDK + CLI): boundary event, summary re-emit, single `session_id`, and `readHistory` of the same session reconstructs `one → two → [marker w/ summary] → three` with no fake user prompt.
- `pnpm check:ci` green; `pnpm test` 568/568 (new: live compaction routing, supplement builder incl. double-compaction, history prepend/dedup, reducer merge).


AprilNEA:
Rebased onto master and pushed two review follow-ups (old head was 091d330).

**Rebase**: master had moved to wire v23 (account schema, PR #71), so the bump here is now **23 → 24** (the root `AGENTS.md` invariant note follows). Only `packages/schema/src/wire/index.ts` conflicted; everything else replayed cleanly. Heads-up: open PR #61 (amp) carries a stale `17 → 18` bump — whichever merges second must re-bump.

**Follow-ups** (from review):
- `dd43068` — `readHistory` now reads the raw-transcript supplement only for the first page (`offset === 0`): the swapped-in summary is the SDK chain's head row and dropped rows are prepended there, so later pages were paying a whole-transcript read + parse for nothing. Covered by a new test (`supplementReads` stays at 1 across a `cursor: '1000'` read). Also noted in a comment why the page-local dedup window is safe.
- `820ebef` — the mobile timeline previously fell through `default: return null` and silently hid the marker; it now renders a compact "Context compacted (pre → post tokens)" row (`mobile.conversation` keys added to both locales). `formatTokens` gained an M tier so a >1M pre-count doesn't read as "1193.4k".

`pnpm check:ci` green, `pnpm test` 618/618 locally on the rebased branch.

PR #77, opened 2026-07-10
chore: monorepo eslint typecheck setup

SukkaW:
- Singleton projectService, no multiple services per lint
- Enable ESLint multi-thread lint
- Enable TypeScript project references, one root typecheck check all


tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    { "path": "packages/client-core" },
+    { "path": "packages/common" },
+    { "path": "packages/engine" },
+    { "path": "packages/i18n" },
+    { "path": "packages/ipc" },
+    { "path": "packages/schema" },
+    { "path": "packages/sdk" },
+    { "path": "packages/transport" },
+    { "path": "packages/ui" },
+    { "path": "packages/workbench" }
```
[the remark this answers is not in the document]
SukkaW: The PR is specifically set up and already works around and avoids TS6306. As you can see, CI is all green.

eslint.config.cjs
```diff
@@ -36,16 +15,24 @@ module.exports = require('eslint-config-sukka').sukka(
         '**/.expo/**',
         '.zed/**',
         '**/expo-export/**',
-        'packages/coss-ui/**',
+        'packages/coss-ui/**', // sync from upstream
+        'packages/sdk/**', // codegen
```
Zerlight: **Do not exclude the hand-written SDK from ESLint.**

`packages/sdk` is not generated code: `docs/ARCHITECTURE.md` explicitly describes it as “Hand-written RPC, not OpenAPI-generated.” This global ignore makes the root `pnpm lint` silently skip production SDK code and also breaks the retained package script:

```text
$ pnpm -F @linkcode/sdk run lint
You are linting ".", but all of the files matching the glob pattern "." are ignored.
Exit status 2
```

On this same head, `pnpm exec eslint --no-ignore packages/sdk` exits 0, so the current SDK sources already satisfy the new configuration. Please remove this ignore. A green root lint must not mean an entire hand-written package was omitted.

eslint-suppressions.json
```diff
@@ -0,0 +1,12 @@
+{
+  "apps/desktop/scripts/verify-artifacts.mts": {
+    "sukka/prefer-foxts-object-size": {
+      "count": 1
+    }
+  },
+  "packages/workbench/src/runtime/provider.tsx": {
+    "sukka/react-no-use-effect-watching": {
+      "count": 1
+    }
```
Zerlight: **Use a cwd-independent, justified suppression.**

ESLint resolves the default `eslint-suppressions.json` relative to `process.cwd()`. The root `pnpm lint` sees this file, but the retained package script runs ESLint from `packages/workbench`, so it does not:

```text
$ pnpm -F @linkcode/workbench run lint
src/runtime/provider.tsx:138:7
error  sukka/react-no-use-effect-watching
Exit status 1
```

The same package command passes on the base commit. This bulk entry also cannot carry the justification required by the repository's lint-suppression policy. Please fix the rule violation or restore a narrowly scoped inline disable with a reason. If bulk suppressions are kept, every package command would need an explicit root `--suppressions-location`, which is the broader and more fragile option.

Zerlight (changes requested):
Requesting changes for two lint regressions that the root CI command does not expose: the hand-written SDK is globally excluded from ESLint, and the new root-relative bulk suppression breaks the retained package-level workbench lint command. Both were reproduced against `e8d6a600`; the TypeScript solution build itself passes.

Zerlight:
### Note on `--concurrency=auto` and the pinned Darwin toolchain

This is informational rather than part of my requested changes, because the crash is not an ESLint correctness bug. The PR exposes a bug in the Node build currently supplied by this repository's pinned Nix/devenv toolchain.

On the exact PR head, the repository-prescribed environment resolves Node `24.15.0` on `aarch64-darwin`. `devenv shell -- pnpm lint` lets ESLint `auto` select 7 workers (`availableParallelism() === 14`) and then emits a very large stream of:

```text
Warning: File descriptor 42 opened in unmanaged mode twice
Warning: File descriptor 42 closed but not opened in unmanaged mode
```

before terminating with:

```text
Assertion failed: (errno == EINTR), function uv__io_poll,
file kqueue.c, line 279.
exit 134
```

There is an exact upstream report: [NixOS/nixpkgs#536039](https://github.com/NixOS/nixpkgs/issues/536039). It identifies the affected combination as Nixpkgs Node 24.15/24.16 on Darwin, `worker_threads`, shared libuv, and libc++ 21. The deeper diagnosis is a V8 `std::hash` ODR violation exposed by the libc++ change ([analysis](https://github.com/NixOS/nixpkgs/issues/536039#issuecomment-4846977926)). The Nixpkgs fixes are merged in [#536043](https://github.com/NixOS/nixpkgs/pull/536043) and [#536336](https://github.com/NixOS/nixpkgs/pull/536336), but the current lock still resolves the affected Node build.

So the attribution matters:

- ESLint concurrency is the trigger because it introduces worker threads; it is not the root cause.
- Linux CI is not evidence that the pinned macOS devenv path is healthy.
- `--concurrency=off` and fixed `--concurrency=2` both completed locally; the latter only reduces exposure and is not the root fix.
- Under an allowed non-Nix Node 26 runtime, all modes complete, but the measured full-lint times were `off: 34.90s`, `2: 32.16s`, and `auto: 67.94s`, so `auto` is also not a demonstrated performance win for this typed-lint workload.

The durable fix is to update the devenv/nixpkgs lock to a revision containing the Darwin Node fix (or move the devshell to Node 26). Until then, disabling or capping concurrency avoids breaking the documented local workflow.

eslint-suppressions.json
SukkaW: Fixed in [`15ad368` (this PR)](https://github.com/arcboxlabs/linkcode/pull/77/commits/15ad368e506e0ceab1f5b340c04e1bb0e2851893)

eslint.config.cjs
SukkaW: Fixed in [`15ad368` (this PR)](https://github.com/arcboxlabs/linkcode/pull/77/commits/15ad368e506e0ceab1f5b340c04e1bb0e2851893)

Zerlight (approved):

PR #101, opened 2026-07-14
fix(desktop): resolve cloud-auth storage path lazily and prune undecryptable entries

AprilNEA:
Fixes CODE-166.

## Problem

The dev shell and the packaged release share one `cloud-auth.json`: `createSafeStorage()` captures `app.getPath('userData')` at module scope (via the auth client's top-level plugin construction), which runs **before** `index.ts` re-points userData to the app identity's directory. The store is therefore pinned to the productName-derived **release** directory even in dev shells. Each side writes entries under its own per-binary safeStorage keychain key, so both apps spam `[cloud-auth] failed to decode stored session data` on every better-auth session refresh (338 hits in the dev log, 218 in the release log since 07-07).

## Fix

- Resolve the store path at call time (`storageFile()`), never at construction — `getItem`/`setItem` only run after identity setup.
- Prune an undecryptable entry instead of returning `null` forever: foreign ciphertext can never heal on its own, so drop it and log once.
- Log `userData: <path>` at boot right after identity setup so this class of drift is visible immediately.

## Verification

- New unit tests (`cloud-auth-storage.test.ts`, following the `settings.test.ts` electron-mock pattern): lazy-resolution regression test (construct → re-point userData → assert the store lands in the new dir), prune-keeps-siblings semantics, plain fallback round-trip.
- Full suite: 102 files / 747 tests pass; scoped biome + eslint + `tsc --build` pass.
- Isolated real launch confirms the boot line prints the post-`setPath` directory.

Cleanup for already-poisoned machines: delete `~/Library/Application Support/LinkCode/cloud-auth.json` and sign in again (both identities re-encrypt under their own keys from then on).

### DIFF
--- a/apps/desktop/src/main/__tests__/cloud-auth-storage.test.ts
+++ b/apps/desktop/src/main/__tests__/cloud-auth-storage.test.ts
@@ -74,21 +74,26 @@ describe('cloud-auth safe storage', () => {
     expect(storage.getItem('session')).toEqual({ token: 't' });
   });
 
-  it('discards an undecryptable entry and keeps the rest', async () => {
+  it('keeps an undecryptable entry on disk and warns once', async () => {
     const storage = await createStorage();
     storage.setItem('good', 'kept');
     mocks.keychainKey = 'B';
-    storage.setItem('foreign', 'poison');
+    storage.setItem('stuck', 'recoverable');
     mocks.keychainKey = 'A';
 
-    expect(storage.getItem('foreign')).toBeNull();
-    expect(mocks.warn).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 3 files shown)


AprilNEA:
Addressed the P2 from both reviewers in aa8b1c1: a decode failure is indistinguishable from a transient keychain failure (locked keychain, declined ACL prompt after a binary change), so pruning could destroy a recoverable session. The entry now survives on disk and the warning is deduped per key per process — which is all the pruning was buying (a foreign entry gets overwritten by the next successful sign-in anyway). The test now asserts the recovery path: decode fails → null + one warning; key accessible again → value comes back.

PR #109, opened 2026-07-14
feat(agent-adapter): opencode history (list/read/resume) via shared history server

lucas77778:
Closes CODE-171 (Phase 2 of CODE-140).

## What

Gives the opencode adapter real history capabilities — `list` / `read` / `resume` — so cold sessions, session switching (transcript seed reads), and history browse/import work for opencode like they do for claude-code/codex. Until now opencode inherited `{list:false, read:false, resume:false}`: its sessions' past content was unrecoverable for any late-attaching or reconnecting client.

## Design (decision locked on CODE-171)

**Shared history-only server.** History calls run on never-started adapter instances (HistoryService's factory), so they can't use any live session's server. A daemon-shared, lazily-spawned, idle-reaped `opencode serve` backs them instead:

- **Self-spawned with a neutral cwd** (`~/.linkcode/opencode-history`): opencode indexes its cwd as the default workspace and the SDK's `createOpencode()` cannot set cwd — a daemon launched from `$HOME` would index the whole home tree.
- **Explicit free port**: `--port=0` does NOT auto-allocate (falls back to 4096, verified live on 1.17.11).
- **SIGTERM→SIGKILL escalation** (the SDK's `close()` is a bare SIGTERM), 60s idle reap, process-exit cleanup, `unref` after readiness so an idle server never holds the daemon's event loop.

Live sessions keep their per-session server; consolidating live + history onto one server (the full paseo model) is a deliberate non-goal here.

## Behavior notes (all live-verified on opencode 1.17.11)

- `readHistory` full-fetches then slices at event level — the messages RPC's `limit` returns the **last** n messages, so it cannot page forward. Reverted messages are truncated at `Session.revert.messageID` (partial `partID` reverts keep everything rather than over-cut).
- Replay reuses the live part ids (`streamDelta` keys / `toolCallFromPart` snapshots) so live turns and cold replay converge by id in the conversation fold.
- `resumeHistory` is a native continue: adopt the existing session id (no `session.create`), scope every call by the session's **own** directory (events ride the per-directory instance bus; the resume cwd need not match).
- Fresh live sessions now emit `session-ref`, **deferred to the first on-stream busy acknowledgement** — announcing at create would seed the client against an empty transcript and the `uptoSeq` cut would swallow the first prompt (same deferral codex uses).

## Verification

- `pnpm check:ci` + `pnpm test` (112 files / 839 tests) green.
- Live drive against a real `opencode serve` + real on-disk sessions: listed 18 sessions sorted desc, replayed a real session as `user-message`/`tool-call`/`agent-message-chunk` events, paged with cursors, resumed a real session (ref announced, no create, correct directory scoping), and confirmed no stray server after dispose.


lucas77778:
Review pass (8-angle parallel scan → 15 candidates → verified 10) + all 10 findings fixed in 4 follow-up commits:

- `f4e0747` hoist `allocatePort` into `@linkcode/common/node` (was duplicated in engine's `scripts/ports.ts` and this PR's history server)
- `005de1a` history-server lifecycle: process-exit hook registered once per manager lifetime (an old generation's exit could previously strip the NEW generation's cleanup via the shared listener reference → orphaned server on daemon exit); readiness detection moved to a rolling window (a >8KB startup preamble could freeze the capped buffer and fake a timeout); one automatic fresh-port retry on a startup-window death (stolen-port race)
- `3943c3c` `canonicalDirectory` for the history cwd filter (symlinked cwds like macOS `/tmp` → `/private/tmp` silently dropped sessions); `readHistory` fetches `get`+`messages` concurrently; shared `FakeEventStream` test double; `textHistoryEvent` reuse in the replay mapping
- `3c8d8b3` hardening from the adversarial re-verify: startup retry honors an in-flight graceful shutdown; path roots survive cwd canonicalization (`C:\` ≠ `C:`)

Verification: 3 independent adversarial re-checks (one ran the new tests against the pre-fix code in a worktree — all three new lifecycle tests fail on the old implementation, pass on the new). `pnpm check:ci` + full `pnpm test` green; live re-drive against a real `opencode serve` including the symlinked-cwd filter case.

PR #143, opened 2026-07-16
feat(agent-adapter): opencode set-model + agent-mode switching (CODE-224)

lucas77778:
## Summary

Phase 3 of CODE-140 (control plane), per CODE-224: the opencode adapter now accepts `set-model` and `set-approval-policy`, using the per-prompt-override mechanism (the pick is resent on every `session.promptAsync` / `session.command` as the `model` / `agent` fields — next-turn semantics, in-flight turns unaffected).

- **`set-model`**: store-then-emit (codex precedent). Rejects refs that aren't `providerID/modelID` (a stored bare id would emit a "successful" `model-update` while prompts silently omit the field), and rejects cross-provider switches when a per-account credential was injected at spawn (injection is spawn-time-only, scoped to one provider). Configured/adopted model is reflected at start.
- **Approval-policy axis = opencode's agent axis**: selectable agents from `app.agents({directory})` (`primary`/`all`, non-hidden) are advertised as policies (claude precedent of `plan` riding this axis); default mirrors the TUI (first primary). Permission posture stays config-driven (CODE-136). `$` shell runs under the selected agent; a failed discovery at start is retried on the first shell command, re-arming the axis.
- **Resume** adopts the Session record's last-used `model`/`agent` unless `StartOptions.model` overrides.
- Both catalog fetches (commands + agents) are best-effort and now run concurrently.

## Verification

- Live-verified against a real `opencode serve` (binary 1.18.2 × SDK 1.17.18): same-session model switch (gpt-5.4 → gpt-5.4-mini) and agent switch (build → plan) both route the very next turn — assistant `providerID`/`modelID`/`agent` readback all flip; Session record tracks last-used model/agent; no switched/ack event fires on the legacy bus. Transcript attached to CODE-224.
- 49 unit tests in `opencode.test.ts` (7 new + 2 validation tests), full suite 1054/1054, `pnpm check:ci` clean.
- Multi-angle code review (line-by-line / removed-behavior / cross-file / reuse / efficiency / conventions) — all confirmed findings fixed in this diff; engine/UI cross-checks came back clean (no prevalidation constraint, attach replay unconditional, approval-policy menu fully data-driven, stub `plan` mode deduped generically).

## Out of scope (follow-ups on CODE-140/CODE-224)

Dynamic model catalog on the wire (desktop model picker stays hidden for opencode until then), `variant` (effort) axis, permission-ruleset presets (superseded by the agent axis).


lucas77778:
Re: Greptile's resume/credential finding (confidence note) — confirmed real, and now fixed at the stack tip in #145 (`092b82c`): a credential-carrying resume without an explicit `StartOptions.model` now pre-reads the session's recorded model off the shared history server *before* the spawn, so the spawn-time provider credential injection covers the provider the resumed turns will target. Unit-tested both ways (injection present with a readable record; graceful fallback to the uninjected spawn when the record read fails). Kept in #145 rather than this branch because the two PRs merge back-to-back and rebasing the stack would churn four commits for a nil master-exposure window.

PR #152, opened 2026-07-17
refactor(daemon): Effect v4 boot orchestration (CODE-244, Phase 1)

AprilNEA:
## Summary

Rewrites the daemon's boot/shutdown orchestration (`apps/daemon/src/index.ts`) as an Effect v4 layer graph, and hardens the identity probe (`runtime.ts`) against timeout misclassification. Engine internals are untouched (Phase 2 is explicitly out of scope until Effect v4 reaches stable/LTS). Decision record in CODE-244.

- **Layers** acquire Shared (config, double-start gate, hub) → Engine → Listeners → RuntimeFile → Uplink; finalizers run LIFO, replacing the hand-maintained `servers[]` / `stopUplink` / `once(shutdown)` choreography.
- **Signals** are handled from the moment the runtime starts (previously registered only after full boot): SIGINT/SIGTERM interrupt the root fiber at any boot phase and unwind exactly the layers acquired. A second signal, or a drain exceeding 10s, force-exits.
- **Exit codes** preserved via custom `runMain` teardown: graceful drain `0`, already-running `3` (`DAEMON_EXIT_ALREADY_RUNNING`), anything else `1` (runMain's default would leak `130` on signals, which the supervisor would count as a crash).
- **Probe hardening**: `probeDaemonIdentity` now retries timeouts (2×) before classifying a port occupant as foreign — a slow-but-live daemon misread as foreign let a second daemon of the same profile hunt past it and split `daemon.db`.
- `effect` + `@effect/platform-node` are exact-pinned betas, devDependencies, bundled via tsup `noExternal` (drizzle-orm pattern) — deploy closures unchanged, bundle 6.2 → 7.4 MB, still a single file.

## Verification

- `pnpm check:ci` (0 errors) and `pnpm test` (1074/1074) green.
- Live smoke against the built bundle in an isolated fake `$HOME`:
  - boot → `GET /linkcode` identity OK; SIGTERM → LIFO teardown, `runtime.json` removed, exit 0;
  - double start → `already running`, exit 3;
  - SIGTERM mid-probe (probe pinned by 5s fake agent shims) → clean rollback in ~6ms, no listener bound, no `runtime.json`, exit 0 — this scenario was a hard kill (or, cold-start, a hang) before.
- Not live-tested: the drain-timeout / second-signal force-exit path (needs an artificially hung `engine.stop`), and a packaged utilityProcess run — flagged in CODE-244 as the remaining acceptance item.

## Notes for review

- Boot semantics are otherwise parity with master, including the probe being awaited before listeners bind — #141 (CODE-225) changes that; whichever lands second absorbs the other (in the layer model the fork is deleting one dependency edge).
- Finalizer-order nuance vs master: `engine.stop()` now runs before the explicit `hub.close()` (both idempotent; master double-closed in the opposite order); uplink now stops before `runtime.json` removal instead of after.
- Effect v4 postdates model training cutoffs; added to the root `AGENTS.md` Never-Guess list with a pointer to the installed `.d.ts` / migration guides.


AprilNEA:
Rebased onto master (04f8ad0b) absorbing #141: `EngineLive` no longer awaits the boot probe — `collect()` stays a pending promise passed to the engine as `agentRuntimesReady`, and the consented-agent refresh loop (#140) rides it. Also folded in a review finding from the effect-ts skill: the ordering-only `RuntimeFile`/`Uplink` tags collapsed into one `Layer.effectDiscard` lifecycle layer (same finalizer order).

Re-verified against the built bundle: with the probe pinned by 5s fake agent shims, listeners now bind at ~150ms (previously blocked for the whole probe), and a SIGTERM while the probe is still pending drains cleanly to exit 0 with runtime.json removed. check:ci 0 errors, 1092/1092 tests.

PR #156, opened 2026-07-17
fix(agent-adapter): probe scans PATH before fallback install locations (CODE-220)

PeronGH:
Fixes CODE-220.

## What

`AgentCliProbe.knownLocations()` now derives candidates from the daemon's own `PATH` first — absolute entries only (relative/empty segments dropped, Windows-style quotes stripped), each joined with `binaryName()` — then appends the per-platform fallback table, deduped. First `--version` vendor-marker-verified candidate wins, so PATH order remains the user's declared precedence.

- **win32**: the fallback table stays empty by design — Windows GUI processes inherit the registry-composed user PATH, and every install channel (winget Links, claude installer's `%USERPROFILE%\.local\bin`, scoop shims) rides PATH. Detection was previously *permanently empty* on Windows.
- **linux**: fallback table gains `/usr/bin` (distro packages, e.g. AUR/Arch).
- `--version` probe timeout 5s → 10s (Defender first-touch scans).

Accepted limitations and the design decision trail are on CODE-220; the user-configured binary-path override follow-up is CODE-250.

## Verification

- New `knownLocations` unit tests (PATH ordering, filtering, dedupe, end-to-end `detect()` through a stubbed PATH entry); all probe tests green.
- `pnpm check:ci` green; `pnpm test` has two failures pre-existing on the base commit (`registry-client` env-specific timeout, `engine-file-suggest` flake) — verified identical with this change stashed.
- Live probe on a real machine (WSL2/Arch): detects `/usr/sbin/claude` 2.1.212 (AUR install, previously invisible to the curated table) in ~220ms.


PeronGH:
Recreating from an ASCII branch name (Linear's auto-slug of the Chinese title was used by mistake).

PR #170, opened 2026-07-17
fix(daemon,agent-adapter,engine,assets): hide win32 console windows on daemon child spawns (CODE-236)

PeronGH:
## Problem

On packaged Windows builds, opening the embedded terminal pops a visible system console window (Windows Terminal / conhost). Root cause: the daemon runs console-less inside Electron's `utilityProcess`, and the PTY sidecar (`linkcode-pty.exe`, a console-subsystem Rust binary) was spawned without `windowsHide: true`, so Windows allocates a fresh visible console (`CREATE_NO_WINDOW` missing).

## Change

Sweep of every daemon-side child-process call — all now pass `windowsHide: true` (12 sites):

- **daemon**: PTY sidecar (the reported popup), aigateway sidecar
- **engine**: `runCommand` (every git call), loop verify-check shell runner
- **agent-adapter**: probe `--version` / claude `auth status` / codex `login status`, claude login, codex app-server, grok headless, opencode history server (cross-spawn)
- **assets**: win32 zip extraction via System32 bsdtar

Also records the rule as root `AGENTS.md` invariant #6 (silent, Windows-only failure mode).

Out of scope (node_modules): claude SDK already hides its own CLI spawns (verified in sdk.mjs); opencode SDK's `createOpencodeServer` does not — noted on the Linear issue.

## Verification

- `pnpm check:ci` clean; targeted vitest run over the touched modules (82 passed; sidecar integration test skipped — needs the built Rust binary)
- **Outstanding**: packaged-Windows manual check (open terminal / codex session / git refresh → no console window) — tracked on CODE-236

https://linear.app/arcbox/issue/CODE-236

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,6 +11,7 @@ Each of these breaks the product, a release, or the build with **no loud error**
 3. **Native deps must be allow-listed.** pnpm blocks install scripts by default; a native dep (e.g. `better-sqlite3`) missing from `allowBuilds:` in `pnpm-workspace.yaml` installs fine but fails at `require()` time with missing bindings.
 4. **CI does not run `pnpm test`.** The vitest suite (~57 test files, node-env pure logic) is absent from `check:ci` and from every workflow — CI gates only format/lint/typecheck plus the Rust job. A green PR does not mean the tests pass; run `pnpm test` yourself before every commit.
 5. **A release is a version+tag pair.** Bump `apps/desktop/package.json` `version`, then push a `v*.*.*` tag; CI fails unless `v${version}` equals the tag. Never hand-tag or hand-edit workspace versions ([`docs/RELEASE.md`](docs/RELEASE.md)).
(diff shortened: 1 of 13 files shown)


PeronGH:
Mechanism verified on real Windows (Node v26.4.0 via WSL interop), reproducing the daemon's exact parent condition — a console-less process (DETACHED_PROCESS, same as Electron `utilityProcess`) spawning a console-subsystem child:

| spawn options | result |
| --- | --- |
| without `windowsHide` (pre-fix) | `GetConsoleWindow()=non-null, IsWindowVisible=true` — visible console window appears (the reported popup) |
| with `windowsHide: true` (this PR) | `GetConsoleWindow()=null, IsWindowVisible=false` — no console window |

Remaining before close: a smoke pass on an actual packaged build (terminal open / codex session / git refresh), tracked on CODE-236.

AprilNEA (approved):

PR #203, opened 2026-07-19
feat(composer): add structured directives and startup controls

Zerlight:
## Summary

- Replace the textarea composer with a Lexical-based editor supporting atomic slash-command, shell-command, and file-mention chips.
- Add keyboard-accessible command menus, directive recovery guidance, attachment conflict handling, and reliable draft submission behavior.
- Carry model and effort selections through new-session startup and persist only provider-confirmed values.
- Align capability, command catalog, model, and effort reflection across agent adapters.
- Bump the wire protocol from 40 to 41 for the startup effort contract.

## Correctness fixes

- Preserve drafts until the host accepts a submission.
- Prevent duplicate submissions and stale file-mention results.
- Treat mid-line slash commands and a bare `$` as regular prose.
- Block incompatible directive and attachment combinations with recovery actions.
- Keep model and effort selections truthful when startup or runtime reflection rejects them.
- Harden Claude Code and OpenCode runtime reflection.

## Verification

- [x] `pnpm check:ci`
- [x] `pnpm test` — 178 test files, 1,457 tests
- [x] Electron mock-runtime smoke test covering command-menu accessibility, model/effort resets, directive parsing, attachment conflicts, and draft clearing

## Runtime note

The wire protocol changes from 40 to 41. The daemon and all clients must be rebuilt and restarted together.

## Follow-up

[CODE-330](https://linear.app/arcbox/issue/CODE-330/fixagent-align-effort-schema-with-provider-specific-capabilities) tracks the pre-existing mismatch between the shared effort schema and provider-specific capabilities.


Zerlight:
@lucas77778 

The backend changes are essential because they turn composer selections into real provider configuration. Without them, the UI could display a model or effort that the agent never received—or continue showing generic 「Effort」 because it cannot determine the effective runtime value.

> Model, effort, and slash-command controls are not presentation-only state. They form a bidirectional contract: the client requests a setting, the adapter applies it using provider-specific semantics, and the backend reports the effective value back to every attached client. Frontend persistence only remembers what to request for the next session; it cannot prove what the provider is actually running.

### Why each backend change matters

1. Initial effort must cross the wire

`effort` was added to `StartOptions` in `packages/schema/src/agent.ts:40`. The engine passes it to the adapter during `session.start`.

Without this:

- A new-session effort selection would remain cosmetic.
- The provider would start with its own default.
- Claude `max` could not work correctly because it must be supplied when constructing the process.
- The session page could show the user’s selection while the backend ran something different.

`BaseAgentAdapter.start()` in `packages/agent-adapter/src/base.ts:82` applies initial effort through the same provider-specific validation hook used by live changes, before starting the provider.

2. The wire version bump is mandatory

Adding `effort` changes a wire variant, so the protocol moved to version 41 in `packages/schema/src/wire/index.ts:36`.

This is not bookkeeping. With mismatched LinkCode peers, frames can connect successfully but then fail validation and be discarded. Without the bump, an older daemon could silently ignore the new startup setting.

3. Only adapters can determine the true effective value

The renderer cannot reliably infer provider defaults because they depend on SDK settings, CLI configuration, model capability, and resume state.

- Claude reads the effective persisted setting and falls back to the documented provider default in `packages/agent-adapter/src/native/claude-code.ts:414`. It later reconciles model-specific downgrades from the running CLI.
- Codex reads the selected model and `defaultReasoningEffort` from `model/list`, then combines that with `thread/start` or `thread/resume` state in `packages/agent-adapter/src/native/codex/adapter.ts:624`. It also consumes `thread/settings/updated` in the same file at line 778.
- Grok reports its verified defaults and rejects unsupported levels instead of silently converting values such as `xhigh` to `high` in `packages/agent-adapter/src/native/grok-build/adapter.ts:209`.
- OpenCode and Pi report the model actually selected by the provider. They do not expose LinkCode’s normalized effort control, so they should not pretend to support it.

This is what prevents a production UI from displaying an unqualified 「Effort」 or a guessed default model.

4. Effective state must survive attachment and navigation

Adapter events can occur during startup, before the conversation page attaches. They can also be missed after reconnecting.

The engine therefore caches `currentModel` and `currentEffort` in `packages/engine/src/engine.ts:82` and replays them during `session.attach` in the same file at line 882.

Without backend replay:

- The new-session page might initially look correct.
- Entering the session page would lose the startup event.
- The composer would revert to 「Default」 or 「Effort」.
- Reconnecting clients could disagree about what the session is running.

A frontend fallback cannot repair a provider event that was emitted before attachment.

5. Slash-command support needs backend capability and validation

Slash support is not equivalent to sending ordinary prompt text across every adapter. Claude parses leading `/`, Codex uses compaction and skill protocol calls, and OpenCode has command RPCs.

The shared capability matrix in `packages/schema/src/agent.ts:98` establishes what each adapter supports before a live catalog arrives.

The engine now distinguishes:

- Unsupported adapter.
- Catalog still loading.
- Authoritative empty catalog.
- Authoritative catalog containing or missing the command.

That logic is in `packages/engine/src/engine.ts:321`. It fixes the first Claude slash command being rejected merely because asynchronous catalog discovery had not completed yet.

6. Mentions are a different boundary

File mentions do not need SDK support. They require:

- The composer to send the currently selected draft or session `cwd`.
- The daemon’s `file.suggest` endpoint to search only that registered workspace.

The recent mention fix was mainly workbench wiring. The backend remains essential as the scoped filesystem authority, but mentions should not be implemented independently in every agent adapter.

### What is not backend-owned

The last-selected effort preference in `packages/workbench/src/surface/new-session-defaults-store.ts:7` is client-local UX state. It answers:

> What should LinkCode request when creating the next session?

The backend answers a different question:

> What model and effort is this running session actually using?

Keeping those separate avoids mutating provider-wide configuration that could also affect Claude Code, Codex CLI, or other applications.

One caveat to communicate: current persistence means 「accepted by LinkCode」, not always strictly 「provider-confirmed」. Codex stores the requested effort before the next turn sends it, while `thread/settings/updated` reconciles it afterward. Strict confirmation would require persisting only after the authoritative provider reflection—ideally by moving Codex effort changes through `thread/settings/update` and tightening initial Claude rejection handling.

AprilNEA (approved):

PR #224, opened 2026-07-20
feat(ui,workbench,daemon): highlight video file paths and play them in the Browser pane

lucas77778:
Closes CODE-316.

Highlights video file paths in chat and plays them in the desktop **Browser pane** (Electron `<webview>`) — deliberately **not** the Files viewer, so playback bypasses `file.read`'s 10 MB cap and streams over HTTP Range (seek, large files, native codecs).

Bumps `WIRE_PROTOCOL_VERSION` 41 → 42 (rebuild + restart daemon and every client together).

## Commits (atomic; each passes repo-wide format + lint + typecheck)

1. **transport** — new `{ filePath, contentType }` `PreviewRoute` variant + Range-capable file streaming in the preview proxy (206/416/HEAD, `Accept-Ranges`, `Content-Range`, `createReadStream`). Unit-tested (`parseByteRange` + a live server: whole-file / partial / 416 / 404).
2. **schema, engine, client** — `file.host`/`file.hosted` wire methods + `HostedFile` model + version bump; `FileHostService` (path-addressed `file--<hash>.localhost`, extension→mime, LRU, shutdown-close) wired through `FileRequestHandler`; SDK `hostWorkspaceFile` + client pending/dispatch plumbing; dev-mock-host placeholder. Unit-tested.
3. **ui** — `video` kind (mp4/m4v/webm/mov/mkv/ogv) in `file-kind.ts`; `openVideoPreview` host action; `artifactNavigationAction` routes video → browser preview (falls back to `openFile`), shared by inline code and file cards. Unit-tested.
4. **desktop** — `onHostVideoFile` on the workbench shell contract (backed by `hostWorkspaceFile`); desktop `openVideoPreview` = locate path → host over daemon HTTP → `openBrowserUrl`.

## Design notes

- **No host-CSP change**: the Browser pane `<webview>` is an independent Chromium context, so the renderer `<meta>` CSP doesn't gate playback. `openBrowserUrl` already navigates the webview to arbitrary loopback URLs.
- **Bypasses the Files viewer + `file.read` 10 MB cap**: video streams over the preview proxy with Range.
- **webview client (`apps/webview`)** has no Browser pane / `openVideoPreview` → video paths fall back to `openFile` (unchanged behavior).

## Testing

- `pnpm test` green (199 files / 1582 tests); repo-wide `pnpm check:ci` passes.
- **Not yet verified** (needs a real desktop+daemon run, can't do headless): actually playing a clip in the Browser pane end-to-end, and codec coverage (H.264/AAC, VP9/webm play; HEVC .mov / some .mkv may surface Chromium's native unsupported-codec error).


packages/foundation/transport/src/preview-proxy.ts
```diff
@@ -48,6 +54,89 @@ function serveContent(res: ServerResponse, body: string, contentType: string): v
...
+    return;
+  }
+
+  const { start, end } = range ?? { start: 0, end: size - 1 };
+  res.writeHead(range ? 206 : 200, {
+    ...commonHeaders,
+    'content-length': String(end - start + 1),
+    ...(range && { 'content-range': `bytes ${start}-${end}/${size}` }),
+  });
+  const stream = createReadStream(filePath, { start, end });
```
Zerlight, line 139: **[P2] Handle zero-length files before creating the stream.** When `size === 0` and no Range header is present, the default range above becomes `{ start: 0, end: -1 }`. `createReadStream(filePath, { start, end })` then throws `ERR_OUT_OF_RANGE` synchronously; because the caller discards the `serveFile` promise, this becomes an unhandled rejection and the HTTP response never completes. I reproduced this with a zero-byte hosted file: the client timed out and Vitest reported the rejection at this line. Return an empty `200` response with `Content-Length: 0` before constructing the stream (and add the regression case).

apps/desktop/src/renderer/src/shell/desktop-shell.tsx
```diff
@@ -312,6 +313,16 @@ export function DesktopShell({
     void locateFileArtifact(path, cwd, conversation.items).then(openRightFileTab);
   }
 
+  // Video plays in the Browser pane: locate the real path, ask the daemon to host it (streamed
+  // with Range), then navigate the in-app browser to that URL — never the 10 MB file viewer.
+  function openVideoPreview(path: string): void {
+    const cwd = active?.cwd;
+    if (!cwd || !onHostVideoFile) return;
+    void locateFileArtifact(path, cwd, conversation.items)
```
Zerlight, line 322: **[P2] Do not resolve video candidates through the 10 MB-capped reader.** `locateFileArtifact` probes candidates with `readWorkspaceFile`, so every video over `MAX_FILE_READ_BYTES` fails the existence probe. With more than one same-name candidate it eventually falls back to `candidates[0]`, which can be stale. I reproduced a latest `/stale/clip.mp4` plus an existing oversized `/actual/clip.mp4`: this chain tried to host the stale path. Resolve/host candidates through a size-independent stat/host operation, or preserve `limit_exceeded` as evidence that the candidate exists.

Zerlight (changes requested):
Two correctness issues need addressing before merge. Both were reproduced against this head: zero-byte hosted files leave the HTTP request hanging with an unhandled `ERR_OUT_OF_RANGE`, and large-video path resolution can select a stale candidate because it probes through the 10 MB-capped `file.read` path. The existing full suite passes, but neither case is covered. Please also complete the repository-required real desktop + daemon playback check after fixing these.

packages/foundation/transport/src/preview-proxy.ts
lucas77778: Fixed in eaff34b1. `serveFile` now returns an empty `200` response with `Content-Length: 0` before constructing a read stream for a zero-length file. Added a regression test that performs the live HTTP request and verifies the empty response completes. Targeted tests, `pnpm check:ci`, and the full 1,584-test suite pass.

apps/desktop/src/renderer/src/shell/desktop-shell.tsx
lucas77778: Fixed in eaff34b1 using the suggested `limit_exceeded` path. `locateFileArtifact` now treats that code as positive existence evidence: `file.read` emits it only after opening/statting the candidate and confirming it is a regular file. The regression test reproduces a stale first candidate followed by an oversized real candidate and verifies the real path wins. I also completed the real desktop + daemon check with a 15.48 MB MP4: the Browser webview loaded it through the daemon, returned HTTP 206 for Range, exposed native controls, and successfully sought and continued playback without a media error.

Zerlight (approved):

PR #234, opened 2026-07-21
fix(agent): align effort schema with provider-specific capabilities

lucas77778:
## Summary

- extend the normalized effort schema with Codex's distinct `ultra` level and advertise per-model effort capabilities/defaults
- parse Codex `model/list` metadata for pre-session and live catalogs, validating startup and live model/effort selections against the selected model
- keep Codex `ultra` separate from Claude's `ultracode`, update effort filtering/fallbacks, and preserve pending picks across stale settings notifications
- bump the wire protocol to v60 (rebased past master' simulator wire bumps)

## Verification

- `pnpm test` (210 files, 1,634 tests)
- `pnpm check:ci`
- `git diff --check`
- verified the `supportedReasoningEfforts` protocol shape against `@openai/codex 0.144.1`

Linear: [CODE-330](https://linear.app/arcbox/issue/CODE-330/fixagent-align-effort-schema-with-provider-specific-capabilities)


packages/host/agent-adapter/src/native/codex/adapter.ts
```diff
@@ -785,13 +868,17 @@ export class CodexAdapter extends BaseAgentAdapter {
...
-          ? effort.data
-          : this.effort === undefined && model
-            ? this.modelDefaultEfforts.get(model)
-            : undefined;
+        const effective =
+          effort.success && effort.data !== 'ultracode'
+            ? effort.data
+            : this.effort === undefined && model
+              ? this.modelOptions.get(model)?.defaultEffort
+              : undefined;
```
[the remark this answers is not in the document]
lucas77778: Fixed in 84ef667. A settings update that conflicts with a pending model is no longer reflected to the UI, and a mismatched effort is likewise suppressed while a newer effort is pending. The adapter can still learn the provider-effective active model internally without replacing the next-turn selection. The regression test now asserts both the UI events and the subsequent turn/start payload.

packages/host/agent-adapter/src/native/codex/adapter.ts
```diff
@@ -583,31 +651,37 @@ export class CodexAdapter extends BaseAgentAdapter {
...
-    this.holdSessionRef = !resume;
     try {
+      const modelCatalog = await this.readModelCatalog(server);
+      if (modelCatalog) {
+        this.catalogDefaultModel = modelCatalog.defaultModel;
+        this.modelOptions = new Map(modelCatalog.models.map((model) => [model.id, model]));
+        this.emitModels(modelCatalog.models);
+      }
+      if (this.effort !== undefined && !resume) {
+        this.assertEffortSupported(this.effort, this.selectedModel());
```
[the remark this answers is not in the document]
lucas77778: Fixed in 84ef667. The adapter now tracks whether the current app-server actually supplied model metadata. If optional model/list discovery is unavailable, max/ultra validation is deferred to Codex instead of blocking thread/start; when a catalog is available, per-model rejection remains enforced. Added parameterized max/ultra regression coverage for the method-not-found path.

packages/host/agent-adapter/src/native/codex/adapter.ts
```diff
@@ -583,31 +654,41 @@ export class CodexAdapter extends BaseAgentAdapter {
...
-      }),
-    };
-    // A fresh thread's rollout holds nothing yet: announcing its history id now would trigger the
-    // clients' transcript seed read, whose uptoSeq cut can swallow the first prompt. Hold the
-    // announcement until the first turn is running; a resumed thread's rollout is complete, so
-    // announce immediately. Set before the request: thread/started can outrun the response.
-    this.holdSessionRef = !resume;
     try {
+      const modelCatalog = await this.readModelCatalog(server);
+      this.modelCatalogAvailable = modelCatalog !== undefined;
```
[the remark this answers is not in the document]
lucas77778: Fixed in 186e6fd. A catalog is now authoritative only when normalization yields at least one usable model entry. Empty or malformed successful responses clear catalog state and defer max/ultra validation to Codex, matching the method-not-found fallback. Added parameterized empty-catalog regression coverage for both levels.

packages/host/agent-adapter/src/native/codex/adapter.ts
```diff
@@ -643,15 +727,22 @@ export class CodexAdapter extends BaseAgentAdapter {
   private reflectThreadSettings(response: unknown, fallbackModel?: string): void {
     if (!isRecord(response)) return;
     const model = stringField(response, 'model') ?? fallbackModel;
-    if (model) this.emitModel(model);
-    if (this.effort !== undefined) return;
+    if (model) {
+      this.activeModel = model;
+      if (this.model !== undefined) this.model = model;
```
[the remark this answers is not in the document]
lucas77778: Fixed in c48a821. On thread/resume, the response model is now recorded only as the provider active model while any explicit next-turn model override is preserved, so recovery cannot reflect or send the stale model. Pending effort validation also uses the preserved model. Added an adapter-level crash/recovery regression that selects Terra + ultra, resumes a thread reporting Luna, and verifies the recovered turn/start still sends Terra + ultra. Full suite: 210 files / 1,639 tests passed; format, lint, and typecheck passed.

packages/host/agent-adapter/src/native/codex/adapter.ts
```diff
@@ -785,13 +881,25 @@ export class CodexAdapter extends BaseAgentAdapter {
         const settings = recordField(params, 'threadSettings');
         if (!settings) break;
         const model = stringField(settings, 'model');
-        if (model) this.emitModel(model);
+        const modelMatchesPending = this.model === undefined || model === this.model;
```
Zerlight: `this.model` is both the sticky turn override and the pending-selection marker, and it is never cleared after a matching settings update. Once a user picks Terra/high, a later authoritative correction to Luna/medium still makes `modelMatchesPending` false forever, so this branch suppresses the provider state and leaves the UI/cache showing values Codex is not using. Keep separate pending model/effort markers and clear them when the provider confirms the pick; the sticky overrides can remain for subsequent `turn/start` requests.

Zerlight (changes requested):
Requesting changes for two correctness issues:

1. The inline comment covers the live pending-selection lifetime.
2. The static fallback capability metadata is not used by draft submission. `new-session-surface.tsx:169-174` resolves `modelOption` only from `dynamicModels`, while this PR adds model-specific fallback capabilities in `agent-models.ts:90-108`. If `model/list` is unavailable, a remembered Sol `ultra` selection can survive after selecting Luna and is submitted even though the fallback table excludes it, causing the first turn to fail. Resolve the selected model from the dynamic catalog or `AGENT_MODEL_OPTIONS[provider]` consistently before calculating `constrainedEffort`.

The pinned 0.144.1 protocol shape and CI both check out; these are the remaining blockers.

packages/host/agent-adapter/src/native/codex/adapter.ts
lucas77778: Fixed in cd7cb81. Sticky model/effort turn overrides are now separate from pending selection markers. Matching settings updates clear each marker independently, and a matching thread/resume model response clears the recovered model marker, so later authoritative provider corrections are reflected instead of being suppressed forever. Unconfirmed picks still survive app-server recovery. Added stale → confirm → correction and resume-confirmation regressions; `pnpm check:ci` and all 1,641 tests pass.

lucas77778:
Addressed the second review blocker in cd7cb81. `NewSessionSurface` now resolves the selected model from the non-empty dynamic catalog or the same `AGENT_MODEL_OPTIONS[provider]` fallback used by `Composer`, through `resolveModel`, before calculating `constrainedEffort`. When `model/list` is unavailable, switching a remembered Sol/ultra draft to Luna now drops the incompatible effort and omits it from submission. Added a regression for that exact flow. `pnpm check:ci` and the full suite (210 files / 1,641 tests) pass.

packages/host/agent-adapter/src/native/codex/adapter.ts
```diff
@@ -667,31 +756,45 @@ export class CodexAdapter extends BaseAgentAdapter {
...
+        this.emitModel(model);
+      } else {
+        this.emitModel(this.pendingModel);
+        if (model === this.pendingModel) this.pendingModel = undefined;
+      }
+    }
+    if (this.effort !== undefined) {
+      this.assertEffortSupported(this.effort, this.model ?? model, !this.modelCatalogAvailable);
+    }
+    if (this.pendingEffort !== undefined) return;
```
Zerlight: **[P1] Clear an effort confirmed by `thread/resume`**

`BaseAgentAdapter.start()` calls `onSetEffort()` before `onStart()`, so resuming with a remembered effort such as `high` sets `pendingEffort` before `thread/resume`. If the resume response already reports `high`, this early return leaves the marker set. The app-server only emits `thread/settings/updated` when the effective settings actually change, so another matching `high` notification is not guaranteed. A later authoritative correction to `medium` then reaches the pending-effort branch and is discarded indefinitely, leaving the engine/UI cache at `high`.

Resolve the response effort (or model default) before returning and clear `pendingEffort` when it matches, paralleling the `pendingModel` handling above. I reproduced this with the existing `TestCodex` fixture: after a resume response with `high`, a `medium` settings update emitted no `effort-update`, both inside and outside the sandbox.

packages/host/agent-adapter/AGENTS.md
```diff
@@ -70,7 +70,7 @@ Every new adapter MUST honor these (`base.ts`); downstream relies on them, they
...
+- **set-model / set-effort** are stored and sent as `turn/start` overrides (they stick for subsequent turns). Effort availability comes from each `model/list` entry's `supportedReasoningEfforts`: Sol/Terra advertise `low|medium|high|xhigh|max|ultra`, while Luna omits `ultra`; `minimal` is outside LinkCode's normalized vocabulary and Claude's distinct `ultracode` always rejects. The adapter publishes that per-model catalog both before session start and on the live event stream, and validates startup/live selections against the selected model. If optional `model/list` discovery is unavailable, validation defers to Codex instead of blocking session start. At startup, it reflects the `thread/start`/`thread/resume` response's model and configured effort; a null effort resolves through that model's `model/list.defaultReasoningEffort` without pinning it as an override. `thread/settings/updated` reflects effective provider state unless it conflicts with a newer pending next-turn pick. Switches apply from the next turn, not mid-turn.
```
Zerlight, line 73: **[P2] Keep the capability matrix in sync**

This paragraph now documents model-specific `max` and `ultra` support, but the capability matrix below still says Codex supports only `low–xhigh`. Please update the Codex row so the package's reference documentation agrees with the adapter and UI.

Zerlight (changes requested):
Requesting changes for one remaining correctness issue and one documentation inconsistency.

The previous model-pending fix and the static fallback fix look correct, but the equivalent resume-confirmation path for `pendingEffort` is still incomplete. A focused adapter regression reproduces the stale reflection deterministically.

Verification on this head: the full test suite passed (244 files, 1,965 tests; 3 files and 5 tests skipped), `pnpm check:ci` passed, `git diff --check` passed, and all GitHub checks are green.

packages/host/agent-adapter/src/native/codex/adapter.ts
lucas77778: Fixed in c19005cc. `reflectThreadSettings` now resolves the resume response effort (including the selected model default) before the pending guard and clears `pendingEffort` when that effective value confirms the remembered pick. A later authoritative `thread/settings/updated` correction is therefore reflected normally, while a mismatched resume response still preserves the newer pending selection. Added a focused resume-high → settings-medium regression. Full suite: 1,970 tests passed (1 skipped); format, full-repo lint, typecheck, and `git diff --check` passed.

packages/host/agent-adapter/AGENTS.md
lucas77778: Fixed in c19005cc. The Codex capability-matrix row now documents `low–max` support and the model-specific Sol/Terra `ultra` capability, matching the detailed contract and adapter/UI behavior.

Zerlight (approved):

PR #241, opened 2026-07-22
fix(desktop): don't persist ephemeral preview-proxy URLs (CODE-373)

lucas77778:
## Summary

- add a shared browser URL predicate for the daemon's ephemeral `<kind>--<id>.localhost` preview namespace
- clear blob and preview-proxy URLs during desktop shell serialization and rehydration while preserving normal HTTPS URLs
- cover blob, file, artifact, service preview, and durable URL persistence behavior

## Testing

- `devenv shell -- pnpm check:ci`
- `devenv shell -- pnpm test` (213 files, 1657 tests)

Linear: https://linear.app/arcbox/issue/CODE-373/fixdesktop-dont-persist-ephemeral-preview-proxy-urls-in-the-browser

### DIFF
--- a/apps/desktop/src/renderer/src/__tests__/shell-state.test.ts
+++ b/apps/desktop/src/renderer/src/__tests__/shell-state.test.ts
@@ -9,6 +9,7 @@ import {
   createRightFileTab,
   createRightTerminalTab,
   DEFAULT_LAYOUT,
+  durableBrowserUrl,
   openFileTabState,
   parsePersistedDesktopShellState,
   RIGHT_PANEL_MAX_SIZE,
@@ -234,7 +235,7 @@ describe('desktop shell state persistence', () => {
       activeSection: 'browser',
       terminal: { tabs: [createRightTerminalTab(), createRightTerminalTab()], activeTabId: null },
       files: { tabs: [fileTab, createRightFileTab('/w/report.pdf')], activeTabId: fileTab.id },
-      browser: { url: 'http://web--app-1a2b3c.localhost:19523' },
+      browser: { url: 'https://example.com' },
     };
     const source: DesktopShellState = {
       sidebarOpen: false,
@@ -259,13 +260,16 @@ describe('desktop shell state persistence', () => {
       '/w/report.pdf',
     ]);
     expect(parsed.rightPanel.files.activeTabId).toBe(parsed.rightPanel.files.tabs[0].id);
-    expect(parsed.rightPanel.browser.url).toBe('http://web--app-1a2b3c.localhost:19523');
+    expect(parsed.rightPanel.browser.url).toBe('https://example.com');
     expect(panelTypes(parsed.bottomPanel)).toEqual(['files']);
   });
 
-  it('drops renderer-scoped blob URLs from persisted browser state', () => {
+  it.each([
+    'blob:http://localhost:5173/expired-preview',
+    'http://file--3e2d018e14777dcb.localhost:19523/',
+  ])('drops ephemeral URL %s from persisted browser state', (url) => {
     const source = createDefaultDesktopShellState();
-    source.rightPanel.browser.url = 'blob:http://localhost:5173/expired-preview';
+    source.rightPanel.browser.url = url;
 
     const serialized = serializeDesktopShellState(source);
     expect(serialized.rightPanel.browserUrl).toBeNull();
@@ -274,10 +278,20 @@ describe('desktop shell state persistence', () => {
       ...serialized,
       rightPanel: {
         ...serialized.rightPanel,
-        browserUrl: 'blob:http://localhost:5173/expired-preview',
+        browserUrl: url,
       },
     });
     expect(parsed.rightPanel.browser.url).toBeNull();
+  });
+
+  it.each([
+    ['blob:http://localhost:5173/expired-preview', null],
+    ['http://file--3e2d018e14777dcb.localhost:19523/', null],
+    ['http://artifact--turn-123.localhost:19523/', null],
+    ['http://web--app-1a2b3c.localhost:19523/', null],
+    ['https://example.com/path', 'https://example.com/path'],
+  ])('filters durable browser URL %s', (url, expected) => {
+    expect(durableBrowserUrl(url)).toBe(expected);
   });
 });
 
--- a/apps/desktop/src/renderer/src/__tests__/shell-state.test.ts
+++ b/apps/desktop/src/renderer/src/__tests__/shell-state.test.ts
@@ -9,7 +9,6 @@ import {
   createRightFileTab,
   createRightTerminalTab,
   DEFAULT_LAYOUT,
-  durableBrowserUrl,
   openFileTabState,
   parsePersistedDesktopShellState,
   RIGHT_PANEL_MAX_SIZE,
@@ -267,6 +266,8 @@ describe('desktop shell state persistence', () => {
   it.each([
     'blob:http://localhost:5173/expired-preview',
     'http://file--3e2d018e14777dcb.localhost:19523/',
+    'http://artifact--turn-123.localhost:19523/',
+    'http://web--app-1a2b3c.localhost:19523/',
   ])('drops ephemeral URL %s from persisted browser state', (url) => {
     const source = createDefaultDesktopShellState();
     source.rightPanel.browser.url = url;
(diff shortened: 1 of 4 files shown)


packages/presentation/ui/src/shell/browser/normalize.ts
```diff
@@ -18,3 +20,15 @@ export function normalizeBrowserUrl(value: string): string {
 export function isAllowedBrowserUrl(url: string): boolean {
   return url.startsWith('http://') || url.startsWith('https://');
 }
+
+/** Whether a URL belongs to the daemon's ephemeral `<kind>--<id>.localhost` preview namespace. */
+export function isPreviewBrowserUrl(url: string): boolean {
```
AprilNEA: Blocking: this is a verbatim copy of `isPreviewOrigin` in `packages/foundation/transport/src/preview-routes.ts`. Please revert this whole file — the function and the two constants added at the top — and import the existing predicate at the desktop call site instead.

Beyond the duplication, the UI package shouldn't own this: the preview namespace is a transport/engine contract, and `packages/presentation/ui` is not allowed to depend on transport, so it has no way to stay in sync with the definition it is mirroring.

packages/presentation/ui/src/shell/browser/index.ts
```diff
@@ -1,3 +1,3 @@
 export type { BrowserPaneProps } from './browser-pane';
 export { BrowserPane } from './browser-pane';
-export { isAllowedBrowserUrl, normalizeBrowserUrl } from './normalize';
+export { isAllowedBrowserUrl, isPreviewBrowserUrl, normalizeBrowserUrl } from './normalize';
```
AprilNEA:
```suggestion
export { isAllowedBrowserUrl, normalizeBrowserUrl } from './normalize';
```

Reverts with `normalize.ts`.

apps/desktop/src/renderer/src/shell/store/model.ts
```diff
@@ -1,3 +1,4 @@
+import { isPreviewBrowserUrl } from '@linkcode/ui/shell/browser';
```
AprilNEA:
```suggestion
import { isPreviewOrigin } from '@linkcode/transport';
```

`@linkcode/transport` is already a dependency of this app and already lands in the renderer bundle through `client-core`, so this is a free swap — no new weight. Import order is unaffected: `@linkcode/transport` still sorts ahead of `@linkcode/ui/*`.

apps/desktop/src/renderer/src/shell/store/model.ts
```diff
@@ -314,8 +315,8 @@ export function serializeDesktopShellState(state: DesktopShellState): PersistedD
   };
 }
 
-function durableBrowserUrl(url: string | null): string | null {
-  return url?.startsWith('blob:') ? null : url;
+export function durableBrowserUrl(url: string | null): string | null {
+  return url?.startsWith('blob:') || (url !== null && isPreviewBrowserUrl(url)) ? null : url;
 }
```
AprilNEA:
```suggestion
function durableBrowserUrl(url: string | null): string | null {
  if (url === null) return null;
  return url.startsWith('blob:') || isPreviewOrigin(url) ? null : url;
}
```

The early return also drops the `url !== null` repetition. Note this removes the `export`, which only existed for the direct unit test — see the test file.

apps/desktop/src/renderer/src/__tests__/shell-state.test.ts
```diff
@@ -9,6 +9,7 @@ import {
   createRightFileTab,
   createRightTerminalTab,
   DEFAULT_LAYOUT,
+  durableBrowserUrl,
```
AprilNEA:
```suggestion
```

No longer needed once the helper goes back to being private.

apps/desktop/src/renderer/src/__tests__/shell-state.test.ts
```diff
@@ -259,13 +260,16 @@ describe('desktop shell state persistence', () => {
...
     expect(parsed.rightPanel.files.activeTabId).toBe(parsed.rightPanel.files.tabs[0].id);
-    expect(parsed.rightPanel.browser.url).toBe('http://web--app-1a2b3c.localhost:19523');
+    expect(parsed.rightPanel.browser.url).toBe('https://example.com');
     expect(panelTypes(parsed.bottomPanel)).toEqual(['files']);
   });
 
-  it('drops renderer-scoped blob URLs from persisted browser state', () => {
+  it.each([
+    'blob:http://localhost:5173/expired-preview',
+    'http://file--3e2d018e14777dcb.localhost:19523/',
```
AprilNEA, line 268:
```suggestion
    'blob:http://localhost:5173/expired-preview',
    'http://file--3e2d018e14777dcb.localhost:19523/',
    'http://artifact--turn-123.localhost:19523/',
    'http://web--app-1a2b3c.localhost:19523/',
```

Folding the other two kinds in here covers the same ground through the public serialize/parse path, so the block below can go.

apps/desktop/src/renderer/src/__tests__/shell-state.test.ts
```diff
@@ -274,11 +278,21 @@ describe('desktop shell state persistence', () => {
...
   });
+
+  it.each([
+    ['blob:http://localhost:5173/expired-preview', null],
+    ['http://file--3e2d018e14777dcb.localhost:19523/', null],
+    ['http://artifact--turn-123.localhost:19523/', null],
+    ['http://web--app-1a2b3c.localhost:19523/', null],
+    ['https://example.com/path', 'https://example.com/path'],
+  ])('filters durable browser URL %s', (url, expected) => {
+    expect(durableBrowserUrl(url)).toBe(expected);
```
AprilNEA: Please drop this block once the cases above cover all four kinds. It is the only reason `durableBrowserUrl` had to be exported, and it re-asserts through a private seam what the round-trip test already proves — the one case it adds that the round-trip doesn't is the durable-URL pass-through, which the `https://example.com` assertion up at line 263 already covers.

apps/desktop/src/renderer/src/__tests__/shell-state.test.ts
```diff
@@ -259,13 +260,16 @@ describe('desktop shell state persistence', () => {
       '/w/report.pdf',
     ]);
     expect(parsed.rightPanel.files.activeTabId).toBe(parsed.rightPanel.files.tabs[0].id);
-    expect(parsed.rightPanel.browser.url).toBe('http://web--app-1a2b3c.localhost:19523');
+    expect(parsed.rightPanel.browser.url).toBe('https://example.com');
```
AprilNEA, line 262: Not a change request — keep this. I traced the old `web--app-1a2b3c.localhost` value: it arrived in the browser pane's first commit as a generic realistic-looking URL, not as an assertion that preview URLs must survive a restart. A durable URL is the right fixture for a round-trip test, so this edit is correcting the fixture rather than overriding an intended contract.

AprilNEA (changes requested):
The behavior is right and it still merges clean on current master. One blocking item: the predicate duplicates one we already have.

`isPreviewBrowserUrl` is a verbatim copy of `isPreviewOrigin` in `packages/foundation/transport/src/preview-routes.ts` — same suffix constant, same label regex, same `try { new URL(url).hostname }`. Importing it instead of re-deriving it costs nothing here: transport re-exports `preview-routes` from its root entry, that entry is browser-safe (the node-only pieces sit behind the `./server` subpath), `apps/desktop` already depends on `@linkcode/transport`, and `client-core` already pulls it into the renderer bundle. It is also the boundary-correct home for the check — `packages/presentation/ui` is not allowed to import transport, so this belongs at the desktop edge rather than as a second copy in the UI package.

That also settles the review bot's P1 about `http://my--app.localhost:3000`: reusing the shared predicate makes the client agree with the daemon's own definition of the preview namespace by construction, instead of maintaining a parallel heuristic that can drift from it.

Things I verified that are fine, so you don't need to revisit them:

- The separate `blob:` check must stay — `new URL('blob:…').hostname` is empty, so `isPreviewOrigin` cannot see it.
- Coverage is complete. `web--`, `file--` and `artifact--` are all registered on demand by the engine, so after a restart the route table is empty and every one of them 404s.
- `BrowserPane` has a `url === null` empty state, so dropping to null does not leave a blank pane.

packages/presentation/ui/src/shell/browser/normalize.ts
```diff
@@ -18,3 +20,15 @@ export function normalizeBrowserUrl(value: string): string {
...
+  try {
+    const { hostname } = new URL(url);
+    if (!hostname.endsWith(LOCALHOST_SUFFIX)) return false;
+    const label = hostname.slice(0, -LOCALHOST_SUFFIX.length);
+    return label.includes('--') && PREVIEW_LABEL_RE.test(label);
+  } catch {
```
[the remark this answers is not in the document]
lucas77778: Addressed in 716afc02 by using transport’s canonical `isPreviewOrigin` rather than maintaining a UI-owned copy. The persistence rule now stays aligned with the daemon’s reserved `<kind>--<id>.localhost` preview namespace by construction.

packages/presentation/ui/src/shell/browser/normalize.ts
lucas77778: Addressed in 716afc02: desktop now imports `isPreviewOrigin` from `@linkcode/transport`; the duplicated presentation helper, constants, and export were removed; `durableBrowserUrl` is private again; and the blob/file/artifact/web coverage now runs through the public serialize/parse path. `pnpm check:ci` and the full 213-file / 1654-test suite pass.

PR #247, opened 2026-07-22
feat(sim): linkcode-sim iOS Simulator sidecar — P0 simctl lifecycle (CODE-392)

AprilNEA:
## Summary

New Rust sidecar `crates/linkcode-sim`: the iOS Simulator host the daemon spawns, P0 scope — public `xcrun simctl` only, no private API.

- Framed stdio protocol mirroring `linkcode-pty` (`[u32 LE total][u8 type][body]`): JSON `REQUEST`/`RESULT` + binary `SCREENSHOT` frames (no base64 for captures). Contract in `crates/linkcode-sim/PROTOCOL.md`; the TS client lands as `@linkcode/sim` (CODE-393).
- Ops: `probe` / `list` / `boot` (waits `bootstatus -b`; already-booted succeeds) / `shutdown` / `install` / `launch` / `terminate` / `openUrl` / `screenshot` (jpeg/png). Per-op deadlines; children killed on timeout.
- Each request runs on its own thread — a slow boot never blocks a screenshot; single writer thread owns stdout.
- Structured errors with stable codes (`xcodeMissing` gates the capability; `simctlFailed`/`timeout`/`invalidRequest`/`io`). Compiles and smoke-tests green on non-macOS.
- Apple tools invoked by absolute path (`/usr/bin/xcrun`): PATH can carry non-Apple xcrun stand-ins (nix xcbuild's breaks SDK resolution).
- CI/packaging: `stage-sidecar.mts` now stages a per-platform crate list; `linkcode-sim` ships in `sidecar/${arch}` extraResources on macOS.

## Verification

- `cargo fmt --check` / `clippy` / `test` clean; `pnpm check:ci` + `pnpm test` (1734) green.
- Protocol smoke tests (all platforms): malformed-request recovery, unknown-frame tolerance, probe shape, EOF shutdown.
- **Real-device acceptance loop** (`cargo test -p linkcode-sim --test device_loop -- --ignored`) passed locally on an iPhone 17 Pro / iOS 26.5 simulator: cold boot → install (fixture UIKit app compiled on the fly) → launch (pid) → screenshot (JPEG magic, >10 KB) → terminate → shutdown.
- `stage-sidecar.mts` run locally: both binaries staged under `apps/desktop/sidecar/arm64/`.


AprilNEA:
Addressed the bot findings in c639e5c7:

- **DEVELOPER_DIR/SDKROOT inheritance** (probe + all commands): new `apple_tool()` factory scrubs both vars from every `/usr/bin/xcrun` / `xcode-select` spawn, so a launcher's foreign SDK selection no longer makes a full Xcode report `xcodeMissing`.
- **Oversized screenshot killing the writer**: the screenshot op now checks the encoded frame against `MAX_FRAME_LEN` and returns a `RESULT` error for just that request; the writer thread additionally survives an over-limit frame (drops it, `InvalidInput` only) instead of tearing down and dropping every later reply.
- **Unbounded request threads**: a counting `InflightGate` caps concurrent workers at 24; over-cap requests park the read loop (backpressure) so a retry burst can't exhaust threads/FDs.
- **EOF leaves children running**: shutdown now drains in-flight workers for up to 3s (quick ops finish, replies flush, their `xcrun` children exit) before abandoning the rest — bounded so a multi-minute `bootstatus` still can't hang shutdown (that device keeps booting server-side in CoreSimulatorService regardless).

PR #249, opened 2026-07-22
feat(sim,engine): @linkcode/sim SDK + SimulatorBackend with per-session device ownership (CODE-393)

AprilNEA:
## Summary

Stacked on #247. The TypeScript layer above the `linkcode-sim` sidecar:

- **`@linkcode/sim`** (new package, `packages/host/sim`): framed-stdio codec mirroring the Rust proto (zod-validated at the boundary), and `SimSidecarClient` — lazy spawn, per-request correlation with outer reply deadlines, crash fail-pending + respawn, EOF shutdown. Binary path is a constructor arg; no packaging probing in the package.
- **engine**: `SimulatorBackend` port (structural, mirrors the `PtyBackend` seam — neither package imports the other) + `SimulatorService`, the policy layer: device↔session ownership (auto-claim, `conflict` on cross-session use), ≤4 devices per session (`limit_exceeded`), 10-min idle reclaim of service-booted devices after their session stops (resuming inside the window keeps the device), user-booted devices never auto-shutdown, engine-shutdown reclaim. Wire (CODE-394) and MCP tools (CODE-395) must route through this service.
- **daemon/desktop**: `resolveSimSidecarPath()` (`LINKCODE_SIM_SIDECAR_PATH` override → dev `target/release` fallback; non-macOS resolves to none), injected as `simulatorBackend` only when configured; the packaged supervisor exports the env from `resourcesPath` (macOS only). `docs/ENVIRONMENT.md` updated.

## Verification

- `pnpm check:ci` + `pnpm test` green (unit: 5 client + 5 codec + 7 registry tests).
- **SDK-driven real-device loop** (`LINKCODE_SIM_E2E=1`, opt-in vitest): cold boot → install (fixture UIKit app) → launch (pid) → screenshot (JPEG >10 KB) → terminate → shutdown, 13s on an iPhone 17 / iOS 26.5 simulator; all devices confirmed shut down afterwards.
- Dev daemon booted under an isolated profile with the wiring live: health endpoint serves, no errors, graceful SIGTERM drain.

PR #251, opened 2026-07-22
feat(schema,engine,client-core): simulator wire contract — availability, device commands, screenshot (CODE-394)

AprilNEA:
## Summary

Stacked on #249. The simulator surface reaches the wire (**WIRE_PROTOCOL_VERSION 43 → 44**):

- **schema**: `simulator.*` variants — `status`/`status.result` (the capability gate: `available:false` is an answer, never an error), `list`/`listed`, `devices.changed` broadcast, and session-scoped commands `boot`/`shutdown`/`install`/`launch`(→`launched` pid)/`terminate`/`open-url`/`screenshot`(→`screenshotted` base64). Model shapes in `model/simulator.ts`.
- **engine**: `SimulatorRequestHandler` routes commands through `SimulatorService` (ownership/cap/reclaim enforced there); after boot/shutdown it re-lists and broadcasts `simulator.devices.changed` best-effort (the engine has no CoreSimulator watcher — its own commands are the only observable change source). `service.status()` caches a successful probe, re-probes failures so installing Xcode heals the capability without a restart.
- **client-core**: typed `simulator*` methods on `LinkCodeClient`/`ControlChannel` via the shared `PendingRegistry` correlation, plus `subscribeSimulatorDevicesChanged`.

Availability is request/reply (`simulator.status`) rather than an unsolicited broadcast — it's connection-constant per host, mirroring `agent-runtime.list`; the dynamic part (device state) is what broadcasts.

## Verification

- Schema contract tests (15 accepted / 3 rejected shapes) + engine wire-injection tests (no-backend unavailable + unsupported, status/list/boot→devices.changed, launch/screenshot replies, cross-session `conflict`). `pnpm check:ci` + `pnpm test` (1772) green.
- **Live acceptance** against a dev daemon over real socket.io: `simulatorStatus` → available with simctl path; 11 devices listed; `boot` → `devices.changed` push shows `iPhone 17: Booted`; screenshot 505 KB JPEG over base64 round-trips and decodes; `shutdown` → second push shows `Shutdown`.

### DIFF
--- a/packages/client/core/src/client.ts
+++ b/packages/client/core/src/client.ts
@@ -38,6 +38,9 @@ import type {
   SessionInfo,
   SessionNotification,
   SessionRecord,
+  SimulatorDevice,
+  SimulatorImageFormat,
+  SimulatorStatus,
   StartOptions,
   TerminalMetadata,
   TerminalReplayEvent,
@@ -103,6 +106,7 @@ export interface AssetSettledEvent {
 type AssetProgressCb = (event: AssetProgressEvent) => void;
 type AssetSettledCb = (event: AssetSettledEvent) => void;
 type AgentRuntimesChangedCb = (runtimes: AgentRuntimes) => void;
+type SimulatorDevicesChangedCb = (devices: SimulatorDevice[]) => void;
 type ConnectionCloseCb = (error: Error) => void;
 
 /** A broadcast about a schedule's or its runs' state — the three `schedule.*` push variants. */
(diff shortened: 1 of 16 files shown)


AprilNEA:
Rebased onto master (which now carries the #247+#266 sidecar/service fixes) and addressed this PR's findings in d5c24663:

- **Reject commands for nonexistent sessions**: the service now takes a `hasSession` predicate (wired to the engine's live-session registry, mirroring `TerminalService`) and `claim()` rejects an unknown session with `not_found` before it can claim a device that would never be released.
- **Failed commands retain claims**: mutating commands run through `withClaim`, which rolls back a claim the failed command *created* (a refreshed pre-existing claim is left intact). Boot is the deliberate exception — it reconciles ownership itself, since a failed boot may still have started the device server-side.
- **Boot ownership / session-release races**: fixed in the service on #266 and inherited here via the rebase.

Added regression tests for session rejection and claim rollback; the wire request-handler test now drives real sessions.

PR #254, opened 2026-07-22
feat(agent-adapter,engine,daemon): simulator MCP tools for every MCP-capable agent (CODE-395)

AprilNEA:
## Summary

Stacked on #251. Agents get the simulator through MCP — the CODE-93 passthrough plus the daemon's built-in endpoint, in one atomic change (the pieces are cross-package-interdependent):

- **MCP passthrough (CODE-93 core)**: `StartOptions.mcpServers` (stdio + http) now reaches claude-code (`options.mcpServers` record), codex (`thread/start` `config` dotted keys `mcp_servers.<name>.*`; an http entry arms `experimental_use_rmcp_client`; http headers have no config key → loud reject), and opencode (`createOpencode` `config.mcp`, http→`remote` / stdio→`local`). pi rejects a non-empty `mcpServers` loudly — and the engine never injects for it.
- **Engine injection**: `SessionStartOptionsResolver` appends the per-session endpoint for `MCP_CAPABLE_AGENT_KINDS` on every start/resume path; `EngineDeps` now takes the daemon-constructed `SimulatorService` (+ `SimulatorMcpProvider`) so the MCP endpoint and the engine share one claims registry; session stop releases claims and tokens.
- **Daemon endpoint**: loopback streamable-HTTP MCP server (`@modelcontextprotocol/sdk`), one session-bound token per session, stateless per-request servers. Tools: `sim_list_devices/boot/shutdown/install/launch/terminate/open_url/screenshot` — screenshot returns MCP **image content**, so the capture lands in the agent's context. Ownership/caps enforced via the shared service; `simulator.activity` broadcasts (new wire variant, still protocol 44) drive the panel's “agent is driving this device” badge, and MCP-path boot/shutdown push `simulator.devices.changed` like the wire handler does.

## Verification

- Unit: adapter mapping (incl. codex header rejection), resolver injection gating (pi excluded, absent capability), endpoint integration over a real MCP client (tool list, image content, cross-session `conflict`, token revocation). `pnpm check:ci` + `pnpm test` (1785) green.
- **Live acceptance** against a dev daemon: real **claude-code** (2.1.216) and **codex** (0.144.1) sessions each autonomously ran `sim_list_devices → sim_boot (iPhone 17) → sim_screenshot → sim_shutdown` through the injected endpoint — activity and devices.changed broadcasts observed client-side. The codex run also live-verifies the `mcp_servers.*` dotted-key + rmcp-client bet. opencode passthrough is SDK-typed and unit-tested; live run pending an opencode install.


AprilNEA:
Rebased onto master (carrying #247/#266/#251) and addressed this PR's findings in 90abf338:

- **Failed starts retained MCP tokens**: `discardFailedStart` now calls `onStopped` when it removes the session, so the simulator MCP endpoint token minted during option resolution is released on a failed start, not leaked until daemon shutdown.
- **Injected name could shadow a user server**: the resolver skips injection when the session's `mcpServers` already contains a server of the injected name, so a user-configured `linkcode-sim` is never overwritten (added a regression test).
- **Unbounded MCP request body**: `readBody` now caps at 4 MiB and destroys the request past it, so a client on the loopback endpoint can't grow it without bound.
- **Wire version**: bumped 44 → **45** for the `simulator.activity` broadcast (Invariant 1).

Reconciled with #251: the daemon builds `SimulatorService` (to share it with the MCP endpoint) and the engine wires the live-session validator onto it via a new `setSessionValidator`.

PR #255, opened 2026-07-22
feat(sim): P1 private-API framebuffer streaming + HID injection, crash-isolated (CODE-396)

AprilNEA:
## Summary

Stacked on #254. The P1 private-API layer for `crates/linkcode-sim`: live framebuffer streaming and touch/button injection on iOS 26, ported from baguette (Apache-2.0; `NOTICE` added). Rust FFI via `objc2` (dynamic ObjC messaging) + `libloading`/`dlsym` (private C symbols) — no Swift.

**What works, verified on real hardware (iPhone 17 Pro / iOS 26.5):**
- **HID injection** — `tap`/`swipe`/`button`, stable (10 taps + 5 swipes + 3 buttons, no crashes). Uses the iOS 26 `IOHIDEvent` digitizer recipe (the 9-arg mouse fn is misread as a Home gesture on 26). Runs in the sidecar main process; wired into the RPC (`probe` now reports `interactive: true`).
- **Framebuffer streaming** — SimulatorKit descriptor enumeration → screen-callback registration → `IOSurface` → JPEG. Delivered live 1206×2622 frames (clock advancing across captures). Key recipe details cracked: XPC-proxy objects need raw `objc_msgSend` (not `objc2`'s verified macro); the callback block needs `BLOCK_HAS_SIGNATURE`; surfaces are recycled buffers so the callback only memcpies pixels (the reader thread encodes); a registration-race gate stops callbacks touching a half-registered descriptor.

**The hard part — crash isolation (engineering best-practice):**
The private framebuffer path intermittently aborts hard (`SIGABRT` "unknown class" inside CoreSimulator's XPC machinery) — uncatchable in-process, and worse on the Xcode 26 `SimStreamProcessor` era. So capture runs in a **disposable worker subprocess**: the sidecar spawns `capture-worker`, reads length-prefixed JPEG frames, and respawns it with backoff on crashes; a crash-loop gives up gracefully. **Input and the simctl lifecycle RPCs stay in the parent and are never taken down by a capture crash.** When the private worker gives up, the stream **degrades to `simctl io screenshot`** — slower, but frames never stop.

New sidecar frames/ops (`PROTOCOL.md`): `STREAM_FRAME` + `tap`/`swipe`/`button`/`streamStart`/`streamStop`.

## Verification

- `cargo fmt`/`clippy`/`test` clean.
- **Live**: input ops over the RPC all succeed; `streamStart` delivers frames (private fast-path when the sim state cooperates — 41 frames/5s observed earlier; simctl fallback when the worker gives up — verified frames keep flowing through a crash-loop while the sidecar stays up).

## Follow-up

The private framebuffer path is fragile on the current Xcode 26 / iOS 26 `SimStreamProcessor` simulators (timing-sensitive hard crashes). It is fully contained by the isolation + graceful-degradation architecture, so it never threatens the sidecar; hardening the private path to reliably hit ≥10fps headless on this Xcode is a contained follow-up (the worker + supervisor are the stable foundation, and baguette upstream is the signal source when Xcode shifts these APIs).


AprilNEA:
Rebased onto master and addressed the findings in f82366b2:

- **Stuck worker blocked stream shutdown**: `CaptureStream::drop` now SIGKILLs the live worker (tracked by pid, which stays valid only while its `Child` is un-reaped) before joining, so a worker parked in `read_exact` after opening a framebuffer without delivering can't hang `streamStop` forever.
- **ABA frame dedup**: the pusher now retains the last-sent `Frame` and compares with `Arc::ptr_eq` instead of a raw address, so a freed+reused allocation can't read as "already sent".
- **Silent worker never triggered fallback**: the pusher tracks time since the last delivered frame and degrades to `simctl` after 3s of silence (not just on `is_dead`), so a worker whose framebuffer registration yields no callbacks still produces frames.
- **streamStart retry shape**: an already-streaming retry now returns `{ streaming, fps, scale, alreadyStreaming }` per PROTOCOL.md, not just `{ alreadyStreaming }`.

Two notes: the JPEG-dict use-after-free was already fixed on this branch (commit 2731e5e5 switched to `kCFType*CallBacks`, which retain the CFNumber). And the TS client/backend/wire consumption of `STREAM_FRAME` is intentionally the stacked follow-up #259 — this PR lands the crash-isolated sidecar capture engine; #259 wires the panel to consume it.

AprilNEA:
Fixed the pid-publication race in 1c9eee7c: after storing the worker pid, `supervise` re-checks `stopped` and kills its own child, so a drop that landed between the spawn and the store (and thus read pid 0) can't leave a silent worker to block `wait()`.

PR #266, opened 2026-07-23
feat(sim,engine): @linkcode/sim SDK + SimulatorBackend with per-session device ownership (CODE-393)

AprilNEA:
Supersedes #249 (auto-closed when its stacked base branch was deleted during the #247 merge). Same branch, now rebased onto master with the review-comment fixes folded in.

`@linkcode/sim` typed sidecar client + engine `SimulatorBackend` port + `SimulatorService` (per-session device ownership, caps, idle reclaim), wired into the daemon.

## Review fixes (from the #249 bot review)

- **Stale sidecar-child events**: every child listener is now scoped to its own child, so an old child's delayed `exit`/`error`/`data` after a crash-respawn no longer tears down (or feeds stale bytes into) the replacement.
- **Write failures waited out the deadline**: a synchronous `writeFrame` failure now fails the request immediately instead of leaving it pending until the reply timeout.
- **Boot/ownership race**: if the owning session stops mid-boot, the booted device is re-tracked for idle reclaim instead of being left running untracked.
- **Reclaim freed the device too early**: the idle reclaim now shuts the device down *before* releasing its claim, closing the window where another session could grab and boot the same udid mid-shutdown. New regression tests cover both races.

CODE-393. Stacked under #251.


AprilNEA:
Addressed the review in abc79e12: boot ownership is now reconciled in a `finally` so a boot that fails after CoreSimulator started the device is still reclaimed; the idle reclaim carries a `reclaiming` flag so a session that resumes mid-shutdown keeps its re-acquired claim; and `foxts` is now declared in @linkcode/sim's dependencies. Added a regression test for the resume-during-reclaim case.

PR #304, opened 2026-07-28
fix(ui,agent-adapter): show nearby rows and line numbers in the inline diff card (CODE-399)

Zerlight:
## Summary

The inline diff card in chat showed a bare list of changed lines — no line numbers, and for
claude-code no surrounding rows at all, so you couldn't tell where in a file an edit landed.
Two independent causes, both fixed here.

**The data never arrived.** claude-code built the card from the tool *input* —
`old_string` / `new_string` verbatim — which is the replaced fragment: no line numbers, no
context. The SDK does hand the adapter `tool_use_result.structuredPatch` on the settle frame
(real hunk offsets, three context lines a side), but `toolUseResultEnvelope` dropped it along
with the other bulk payloads. The adapter now lifts it into the `patch` field that already
existed, on both the live and history-replay paths.

**The renderer discarded what it did get.** `patchLines` skipped every `@@` header — the only
place file line numbers exist — so even codex, which already shipped a real git patch, rendered
unnumbered and unhunked. `DiffBlock` now renders through `@pierre/diffs`, the same viewer
already behind the right-panel git diff and the Files viewer: numbered gutter, hunk separators,
context rows, syntax highlighting.

Also in here:

- File cards clamp a long body to a peek with an Expand / Collapse control in the frame footer.
  The control only appears when the body actually overflows, so short diffs are untouched.
- A shared `unifiedPatchText` helper in `@linkcode/schema`, replacing a hand-rolled copy in
  `im-render`.
- Root vitest gains a setup file polyfilling `ResizeObserver` and `Element.prototype.getAnimations`;
  jsdom has neither and pierre / base-ui need both.

Decisions worth flagging for review:

- **Write keeps whole-file `newText`, no patch.** Its dominant `create` case ships an empty
  `structuredPatch`, and for a create the whole file beats an all-`+` patch.
- **`oldText` / `newText` stay alongside `patch`,** as codex already does: consumers prefer
  `patch` when present, and `file-tool-presentation` derives "this was a deletion" from the text
  fields, so dropping them would silently change navigation.
- **Headerless patches are normalized in the UI.** Codex and the new claude path both emit bare
  `@@` hunk streams, and pierre parses zero hunks out of those — the card would render blank —
  so `DiffBlock` synthesizes a `--- path` / `+++ path` preamble.
- **No `WIRE_PROTOCOL_VERSION` bump.** `patch: { format: 'git_patch', text }` already existed
  (CODE-388); this adds a plain TS interface and a function, no zod variant changed, nothing
  under `src/wire` touched.

## Verification

`pnpm check:ci` and `pnpm test` both pass (1967 passed, 1 skipped).

New coverage: the `chatFileDiff` branches (headerless codex patch, multi-hunk git patch,
fragment-only text, create / delete / rename, binary), the peek overflow gate, the adapter's
`editResultDiffContent` extractor, and live-vs-history shape parity for the replayed patch.

Rendering was driven through headless playwright against `dev:mock` rather than the in-app
preview pane — that pane runs `document.hidden`, which freezes the frame loop, so
`ResizeObserver` never delivers and the overflow gate reads as broken when it isn't. In a real
frame loop: a multi-hunk fixture renders absolute line numbers (12–19, then `@@ -84,9 +85,11 @@`,
then 85–90) with context on both sides; the overflowing card gets the clamp, fade and toggle,
and the short one correctly gets none of them.

**Not verified: the history-replay half against a real daemon.** The mock host rejects
`history.read`, so cold-resume parity for a claude Edit was only checked at unit level. Worth a
manual pass before merge — make an Edit, reload the thread, confirm the replayed card matches
the live one.

## Checklist

- [x] I ran the affected surface and observed the change working — mock renderer; see the caveat above
- [x] If a wire message changed: `WIRE_PROTOCOL_VERSION` is bumped — n/a, no wire variant changed

### DIFF
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -47,6 +47,7 @@ module.exports = require('eslint-config-sukka').sukka(
         // Workspace tests belong to their source tsconfig or a tests/tsconfig.json referenced by
         // the root solution; listing them here too would breach typescript-eslint's 8-file cap.
         'vitest.config.ts',
+        'vitest.setup.ts',
       ],
     },
   },
--- a/packages/client/workbench/src/mock/data/showcase.ts
+++ b/packages/client/workbench/src/mock/data/showcase.ts
@@ -436,6 +436,51 @@ export function createShowcaseToolBursts(terminalId = SHOWCASE_TERMINAL_ID): Sho
         ],
       },
       {
+        // The gutter/context case: a real multi-hunk patch, so the card shows absolute line
+        // numbers, surrounding rows, and a `@@` separator between hunks. Every other diff fixture
+        // here is a 1-3 line oldText/newText pair, which exercises none of that.
+        toolCallId: 'mock-tool-edit-multi-hunk',
+        title: 'Rewire the conversation pipeline',
+        kind: 'edit',
+        status: 'completed',
+        content: [
+          {
+            type: 'diff',
+            change: 'modify',
+            path: 'packages/client/core/src/conversation.ts',
+            patch: {
+              format: 'git_patch',
+              text: [
+                '@@ -12,7 +12,8 @@',
+                ' ',
+                ' const MAX_ITEMS = 500;',
+                ' ',
+                '-export function foldTurn(items: ConversationItem[]): Turn {',
+                '+/** Folds one turn, dropping items past the cap. */',
+                '+export function foldTurn(items: readonly ConversationItem[]): Turn {',
+                '   const turn = createTurn();',
+                '   for (const item of items) {',
+                '     turn.push(item);',
+                '@@ -84,9 +85,11 @@',
+                '   return turn;',
+                ' }',
+                ' ',
+                '-function createTurn(): Turn {',
+                '-  return { items: [], tokens: 0 };',
+                '+function createTurn(seed?: Partial<Turn>): Turn {',
+                '+  return { items: [], tokens: 0, ...seed };',
+                ' }',
+                ' ',
+                ' export function isBoundary(item: ConversationItem): boolean {',
+                '-  return item.kind === "user";',
+                '+  return item.kind === "user" || item.kind === "compaction";',
+                ' }',
+              ].join('\n'),
+            },
+          },
+        ],
+      },
+      {
         toolCallId: 'mock-tool-write-plan',
         title: 'Write PLAN.md',
         kind: 'edit',
--- a/packages/foundation/schema/src/model/__tests__/tool-call.test.ts
+++ b/packages/foundation/schema/src/model/__tests__/tool-call.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { ToolCallContentSchema } from '../tool-call';
+import { ToolCallContentSchema, unifiedPatchText } from '../tool-call';
 
 describe('ToolCallContentSchema diff', () => {
   it('keeps legacy oldText/newText diffs readable', () => {
@@ -23,5 +23,20 @@ describe('ToolCallContentSchema diff', () => {
     },
   ])('accepts a structured $change diff without dual full text', (diff) => {
     expect(ToolCallContentSchema.safeParse({ type: 'diff', ...diff }).success).toBe(true);
+  });
+});
+
+describe('unifiedPatchText', () => {
+  it('emits a header per hunk followed by that hunk body', () => {
+    expect(
+      unifiedPatchText([
+        { oldStart: 1, oldLines: 2, newStart: 1, newLines: 2, lines: [' same', '-before'] },
+        { oldStart: 40, oldLines: 1, newStart: 40, newLines: 2, lines: ['+added', ' tail'] },
+      ]),
+    ).toBe('@@ -1,2 +1,2 @@\n same\n-before\n@@ -40,1 +40,2 @@\n+added\n tail');
+  });
+
+  it('renders no hunks as empty text', () => {
+    expect(unifiedPatchText([])).toBe('');
   });
 });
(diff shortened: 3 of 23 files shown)


lucas77778 (changes requested):
## Request changes

The overall direction is sound, but the blocking issue below should be fixed before merge.

### [P1] A panel mounted after the first commit never registers its ResizeObserver

`packages/presentation/ui/src/chat/file-preview-card.tsx:57-70,115-147`

`subscribeToPanelSize` is wrapped in a `useCallback` with no dependencies and reads `panelRef.current` only when the subscription is created. A live Read first mounts as a header-only card, so the ref is null and the subscription returns `noop`. When the settle later mounts the body panel on the same card instance, the subscribe function is unchanged and `useSyncExternalStore` does not resubscribe. The long body is therefore clipped by `chat-card-peek` while `panelOverflowing` remains permanently false, leaving no fade and no Expand button.

This was reproduced against the PR head by rendering the header-only card first, then rerendering it with a body whose `scrollHeight=600` and `clientHeight=200`: the ResizeObserver constructor count remained 0 and the Expand button was absent. Store the panel element through a callback ref and make the element a dependency of the subscribe/getSnapshot callbacks. Please also add a header-only → late-body regression test.

### [P2] Subagent Edit history replay does not recover structuredPatch

`packages/host/agent-adapter/src/native/claude-code.ts:1643-1655`

The main-agent mapper receives `supplement.toolUsePatches`, but `readSubagentTranscripts` still calls `createClaudeHistoryEventMapper(asHistoryId(sessionId))` without a patch supplement. The SDK's `getSubagentMessages` projection drops `toolUseResult`, so a subagent Edit has real hunks and line numbers while live, then falls back to the announce-time old/new fragment after reload. Build a patch supplement from each raw subagent transcript and add a live/history parity test.

### [P2] Keyboard focus can enter visually clipped Markdown controls

`packages/presentation/ui/src/styles.css:346-361`
`packages/presentation/ui/src/chat/file-preview-card.tsx:117-147`

`overflow: clip` hides content visually but does not remove links, checkboxes, or other controls from the tab order, and the Expand button comes after the panel. Keyboard users can therefore focus an invisible control, while `clip` prevents scrolling it into view. Make the collapsed and overflowing panel inert until it is expanded, or automatically expand when focus enters the panel, and add keyboard coverage.

### [P3] Diff statistics disagree with the rendered text fallback

`packages/presentation/ui/src/chat/diff-block.tsx:69-86`
`packages/presentation/ui/src/chat/diff-utils.ts:102-125`

`chatFileDiff` falls back to `oldText`/`newText` when a patch produces no hunks, but `diffContentStats` still uses the empty or invalid patch and reports 0/0. The card can show a visible diff with no addition/deletion badge. When `patchLines` is empty and text is available, `diffStats` should fall back to `diffLines` as well.

Remote CI is currently green and `git diff --check` passes; these cases are not covered by the existing tests.

Zerlight:
> ## Request changes
> The overall direction is sound, but the blocking issue below should be fixed before merge.
> 
> ### [P1] A panel mounted after the first commit never registers its ResizeObserver
> `packages/presentation/ui/src/chat/file-preview-card.tsx:57-70,115-147`
> 
> `subscribeToPanelSize` is wrapped in a `useCallback` with no dependencies and reads `panelRef.current` only when the subscription is created. A live Read first mounts as a header-only card, so the ref is null and the subscription returns `noop`. When the settle later mounts the body panel on the same card instance, the subscribe function is unchanged and `useSyncExternalStore` does not resubscribe. The long body is therefore clipped by `chat-card-peek` while `panelOverflowing` remains permanently false, leaving no fade and no Expand button.
> 
> This was reproduced against the PR head by rendering the header-only card first, then rerendering it with a body whose `scrollHeight=600` and `clientHeight=200`: the ResizeObserver constructor count remained 0 and the Expand button was absent. Store the panel element through a callback ref and make the element a dependency of the subscribe/getSnapshot callbacks. Please also add a header-only → late-body regression test.
> 
> ### [P2] Subagent Edit history replay does not recover structuredPatch
> `packages/host/agent-adapter/src/native/claude-code.ts:1643-1655`
> 
> The main-agent mapper receives `supplement.toolUsePatches`, but `readSubagentTranscripts` still calls `createClaudeHistoryEventMapper(asHistoryId(sessionId))` without a patch supplement. The SDK's `getSubagentMessages` projection drops `toolUseResult`, so a subagent Edit has real hunks and line numbers while live, then falls back to the announce-time old/new fragment after reload. Build a patch supplement from each raw subagent transcript and add a live/history parity test.
> 
> ### [P2] Keyboard focus can enter visually clipped Markdown controls
> `packages/presentation/ui/src/styles.css:346-361` `packages/presentation/ui/src/chat/file-preview-card.tsx:117-147`
> 
> `overflow: clip` hides content visually but does not remove links, checkboxes, or other controls from the tab order, and the Expand button comes after the panel. Keyboard users can therefore focus an invisible control, while `clip` prevents scrolling it into view. Make the collapsed and overflowing panel inert until it is expanded, or automatically expand when focus enters the panel, and add keyboard coverage.
> 
> ### [P3] Diff statistics disagree with the rendered text fallback
> `packages/presentation/ui/src/chat/diff-block.tsx:69-86` `packages/presentation/ui/src/chat/diff-utils.ts:102-125`
> 
> `chatFileDiff` falls back to `oldText`/`newText` when a patch produces no hunks, but `diffContentStats` still uses the empty or invalid patch and reports 0/0. The card can show a visible diff with no addition/deletion badge. When `patchLines` is empty and text is available, `diffStats` should fall back to `diffLines` as well.
> 
> Remote CI is currently green and `git diff --check` passes; these cases are not covered by the existing tests.

Fixed in [9be62c3](https://github.com/arcboxlabs/linkcode/pull/304/commits/9be62c30180a9f1cf29d3218f9ba9ef5dec04f6c) and [1466754](https://github.com/arcboxlabs/linkcode/pull/304/commits/1466754230d4464812cc10b692d91edad6202787).

Some details regarding `[P2] Subagent Edit history replay does not recover structuredPatch` can be found on Linear. This is caused by a upstream issue. See [claude-code #75283](https://github.com/anthropics/claude-code/issues/75283).

lucas77778 (approved):

PR #309, opened 2026-07-29
fix(mobile): polyfill AbortSignal#throwIfAborted for Hermes (CODE-466)

lucas77778:
## Summary

- Polyfill `AbortSignal.prototype.throwIfAborted` for Hermes in the mobile runtime.
- Load the polyfill from the Expo Router root layout before the connection runtime is used.
- Preserve a future native implementation and cover the fallback behavior with unit tests.

`foxts/async-retry` calls `throwIfAborted()` before each `ConnectionController` recovery attempt. Hermes in React Native 0.86 does not provide that method, so mobile host connections failed synchronously before dialing (CODE-466).

## Verification

Local verification requested by `lucas77778`, on Node 26.3.0 after merging current `master`:

- `pnpm format:check` — passed.
- `pnpm lint:ci` with the CI-equivalent single-threaded lint configuration — passed with no errors.
- `pnpm typecheck` — passed.
- `pnpm test` — passed: 257 files passed, 3 skipped; 2,076 tests passed, 5 skipped.
- `pnpm -F @linkcode/mobile smoke:export` — passed for Android and iOS; both outputs were Hermes bytecode bundles and included the required Expo Router modules.
- The earlier CI `Mobile Native Bundles`, desktop, webview, Rust, and CodeQL jobs passed.

The original CI TypeScript job stopped in ESLint with a worker heap OOM before typecheck/tests. Current `master` was merged in commit `3cc966bd`, bringing in the CI fix that uses a 6144 MB heap and single-threaded `pnpm lint:ci`; the push triggered a fresh CI run.

Runtime device verification remains outstanding: the Linux orb has no Android emulator/ADB and cannot run the iOS simulator, so a real Hermes connection to a host was not claimed.

## Checklist

- [ ] `pnpm check:ci` and `pnpm test` both pass (plus `cargo fmt` / `clippy` / `test` for Rust changes)
  - `pnpm test` and the current CI gate (`format:check`, `lint:ci`, `typecheck`) passed; the local convenience `check:ci` still invokes concurrent lint and OOMs in this orb.
- [ ] The affected surface was run and the change observed working
  - Android/iOS production exports passed; device-level host connection verification is still required.
- [x] If a wire message changed: `WIRE_PROTOCOL_VERSION` is bumped (not applicable; no wire change)
- [x] New code and assets are original work, or their origin and license compatibility are noted above

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -1,3 +1,5 @@
+import '../runtime/polyfills';
+
 import { BottomSheetModalProvider } from '@gorhom/bottom-sheet';
 import { sanitizeSentryTransaction } from '@linkcode/common/sentry';
 import type { TelemetryConfig } from '@linkcode/common/telemetry-config';
--- a/apps/mobile/src/runtime/__tests__/polyfills.test.ts
+++ b/apps/mobile/src/runtime/__tests__/polyfills.test.ts
@@ -1,0 +1,42 @@
+import { afterEach, expect, it } from 'vitest';
+import { installPolyfills } from '../polyfills';
+
+// Held as a property descriptor: a bare method reference trips @typescript-eslint/unbound-method.
+const native = Object.getOwnPropertyDescriptor(AbortSignal.prototype, 'throwIfAborted');
+
+afterEach(() => {
+  if (native) Object.defineProperty(AbortSignal.prototype, 'throwIfAborted', native);
+});
+
+it('installs throwIfAborted when the runtime lacks it (Hermes, CODE-466)', () => {
+  // @ts-expect-error -- simulate Hermes, which has no throwIfAborted
+  delete AbortSignal.prototype.throwIfAborted;
+  installPolyfills();
+
+  const fresh = new AbortController();
+  expect(() => fresh.signal.throwIfAborted()).not.toThrow();
+
+  const aborted = new AbortController();
+  const reason = new Error('stop');
+  aborted.abort(reason);
+  expect(() => aborted.signal.throwIfAborted()).toThrow(reason);
+});
+
+it('throws even when the signal was aborted without a reason', () => {
+  // @ts-expect-error -- simulate Hermes, which has no throwIfAborted
+  delete AbortSignal.prototype.throwIfAborted;
+  installPolyfills();
+
+  const controller = new AbortController();
+  controller.abort();
+  // Node fills in an AbortError reason; the polyfill's own fallback covers a runtime that leaves
+  // `reason` undefined. Either way the call must throw.
+  expect(() => controller.signal.throwIfAborted()).toThrow();
+});
+
+it('leaves a native implementation in place', () => {
+  installPolyfills();
+  expect(Object.getOwnPropertyDescriptor(AbortSignal.prototype, 'throwIfAborted')?.value).toBe(
+    native?.value,
+  );
+});
--- a/apps/mobile/src/runtime/__tests__/polyfills.test.ts
+++ b/apps/mobile/src/runtime/__tests__/polyfills.test.ts
@@ -22,16 +22,27 @@ it('installs throwIfAborted when the runtime lacks it (Hermes, CODE-466)', () =>
   expect(() => aborted.signal.throwIfAborted()).toThrow(reason);
 });
 
-it('throws even when the signal was aborted without a reason', () => {
+it('falls back to an AbortError when the signal was aborted without a reason', () => {
   // @ts-expect-error -- simulate Hermes, which has no throwIfAborted
   delete AbortSignal.prototype.throwIfAborted;
   installPolyfills();
 
   const controller = new AbortController();
   controller.abort();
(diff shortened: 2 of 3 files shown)


AprilNEA:
Pushed `95db9640` on top of this branch — three refinements to the same polyfill, no new surface.

**1. The fallback now throws an `AbortError`.** `Error("Aborted")` → `DOMException("The operation was aborted.", "AbortError")`. On RN 0.86 that fallback is not an edge case, it is the *only* path: the injected `abort-controller` has no `reason` property at all, so every abort surfaced as a generic `Error` and anything branching on `name === "AbortError"` would have missed it. The wording matches Expo's own winter-runtime fallback (`expo/src/winter/AbortSignal.ts` → `getAbortReason`), so the two patches agree on what an aborted signal throws. `DOMException` is installed as a global by that same winter runtime, so it is always present here.

**2. The unit test now actually reaches that branch.** Node populates `reason` with its own `AbortError`, so the previous "aborted without a reason" case went through the native reason and never exercised the fallback. The test strips `reason` off the signal first; the message assertion is what pins it, since Node's own reason reads `This operation was aborted` while ours reads `The operation was aborted.`

**3. Corrected the root-cause comment.** The Abort API is not Hermes' — it is a Web API that RN injects. RN 0.86 still depends on `abort-controller@^3.0.0` (2019), which has no `throwIfAborted` and discards the argument to `abort(reason)`; the `AbortSignal.any` the issue probed comes from Expo's winter runtime, which fills in only the `timeout`/`any` statics. Same failure, but the old wording would send the next reader looking in the wrong runtime.

**Removal is tracked as CODE-477.** RN 0.87 replaces the vendored dependency with a built-in Abort API (`src/private/webapis/dom/abort-api`, with `reason` + `throwIfAborted`), so this file becomes dead weight — but npm `latest` is still `0.86.2` (`next` = `0.87.0-rc.3`), and Expo binds exactly one RN version per SDK, so it has to wait for the SDK 58 bump. CODE-477 also records the shorter path: upstreaming `throwIfAborted` into Expo's winter runtime is a JS-only change that could ship in a `57.0.x` patch and retire this file inside the current SDK cycle.

Re-verified locally after the commit: `pnpm format:check`, `pnpm lint` (0 errors), `pnpm typecheck`, and `pnpm test` (257 files / 2076 tests passed) — plus the pre-commit hooks over the whole repo. Device verification of the CODE-466 acceptance path is still outstanding, unchanged by this commit.

AprilNEA:
⚠️ Before this goes any further: **master already carries this fix**, and landing both would be a regression.

Merging current master into this branch conflicts on `apps/mobile/src/app/_layout.tsx:1`. The cause is CODE-462, which landed the same fix through #310 (`33177b7d` → `apps/mobile/src/polyfills.ts`, documented in the `src/polyfills.ts` bullet of `apps/mobile/AGENTS.md`). That commit is not an ancestor of this branch — the master merge here (`3cc966bd`) predates it — which is why the duplication was invisible from inside this PR.

Master's module is a superset of this one: the same guarded `throwIfAborted` install, **plus** a wrapper around `AbortController.prototype.abort` that records the reason under a symbol and an `AbortSignal.prototype.reason` getter that reads it back, with `AbortError` from `foxts/abort-error` as the fallback. That second half is load-bearing on RN 0.86, which discards the argument to `abort(reason)` outright while `ConnectionController` re-throws `signal.reason` when a run is superseded.

**Why both is worse than either.** Master gates the `abort` wrapper, the `reason` getter, and `throwIfAborted` behind one `if (typeof AbortSignal.prototype.throwIfAborted !== 'function')`. If a conflict resolution puts this branch's `import '../runtime/polyfills'` ahead of `import '../polyfills'`, this module installs `throwIfAborted` first, master's guard then sees a function and skips its entire block — and `signal.reason` silently reverts to `undefined`. Nothing throws; the failure would resurface later as an unclassifiable `catch`.

**Suggested disposition:** close this PR as superseded and CODE-466 as a duplicate of CODE-462, instead of resolving the conflict. Deleting the surviving polyfill once RN ships a real Abort API (RN 0.87 / Expo SDK 58, ~September) is tracked in CODE-477.

**One thing here is worth salvaging:** master's `src/polyfills.ts` has no test — `apps/mobile/src/runtime/__tests__/` covers only `startup` and `use-session-actions`. The test from this branch ports over with small changes (drive the module through `vi.resetModules()` + dynamic import, since master patches at module scope rather than through an exported `installPolyfills()`), and it can also pin the `reason` round-trip that only master's version provides. Happy to open that as a separate small PR against master if wanted.

For the record, my commit `95db9640` on this branch was written before I saw #310, so treat it as moot — master's fallback already throws a `foxts` `AbortError`, which is the better shape.

AprilNEA:
Closing as superseded: CODE-462 landed the same fix on master via #310 (`apps/mobile/src/polyfills.ts`), with a superset of this implementation — see the analysis above. CODE-466 is marked a duplicate of CODE-462. The one gap worth keeping from this branch, a unit test for the surviving polyfill, is being ported to master separately; removal of the polyfill on the RN 0.87 / Expo SDK 58 bump is tracked in CODE-477. Branch left in place rather than deleted.

PR #325, opened 2026-07-29
feat(client-core,mobile): narrow event delivery to attached sessions (CODE-446)

AprilNEA:
Closes CODE-446.

Every new connection defaults to subscription mode `all`, so a client is delivered `agent.event` for
every session on the host. Free on desktop over loopback; on a phone over the tunnel it is metered
bytes and battery for threads the user never opened. `subscription.set` was already answered by the
Hub but had no client call sites at all.

## What changed

- `client-core` exposes the three frames the Hub already understood: `setSubscriptionMode` (acked
  request), plus fire-and-forget `detachSession` alongside the existing `attachSession`.
- The mobile conversation screen announces its session on open and withdraws on teardown, and the
  connection asks for `attached` scope once it is ready — re-sent per generation, since a recovered
  connection starts back at `all`.
- `SessionSubscriptionMode` is hoisted out of the inline `z.enum` in the `subscription.set` variant
  so both sides name one type. **No `WIRE_PROTOCOL_VERSION` bump**: the validated shape is
  unchanged, only its spelling.

## On the ordering hazard in the issue

The issue expected `session.attach` to have to happen *after* the seed samples `uptoSeq`, on the
theory that a re-broadcast event landing inside the cut is "neither in the history seed nor past the
cursor, so silently lost". That turns out not to hold, and attach-before-seed is the safe order.

`conversation-store.ts` cuts an event inside `uptoSeq` only when `coveredBySeed` proves the snapshot
contains it, and everything ephemeral — interactive asks, status, policy advertisement — returns
`false` there by construction, because `history.read` can never contain it (CODE-35). So a replayed
approval survives the cut. Attaching first is also strictly better: it is what makes the events
arrive at all under `attached`.

That invariant is now pinned by a test rather than left to reasoning — flipping `coveredBySeed`'s
default to `true` fails it.

## Verification

Acceptance driven against a **live daemon** with two real `claude-code` sessions and two clients,
one scoped to `attached` and attached to one thread only:

| client | open thread | closed thread |
| --- | --- | --- |
| unscoped (`all`) | 10 events | 6 events |
| scoped (`attached`) | 5 events | **0 events** |
| scoped, after attaching the closed thread | — | 0 → **5 replayed** |

`subscription.set('attached')` acked `{ok:true}` by the real daemon; delivery narrowed as specified;
attaching later replayed the buffered state, which is the approval-appears-on-open path.

Automated coverage added on top: the client frames (`control-client.test.ts`), and the mobile hook's
attach/detach plus attach-precedes-read ordering and the ephemeral-ask-survives-the-cut regression
(`use-seeded-conversation.test.ts`).

`pnpm check:ci` and `pnpm test` (2093 passed) green.

## Merge note

This branch predated master's own fix for CODE-444 (the two-React vitest problem). Master's
`apps/mobile/vitest.config.ts` project won; this branch's competing root-config version and its
duplicate `use-session-actions` test were dropped. Master's `ControlledTransport` was extracted to
`__tests__/client-test-helpers.tsx` so the new test could assert wire payloads the same way, which
left `use-session-actions.test.tsx` JSX-free and renamed to `.ts`. `use-terminal-session.test.ts`
came along from that superseded commit — it is unrelated to this issue but is real coverage master
does not have, so it is kept rather than thrown away.


AprilNEA:
Both P1 findings confirmed against the code and fixed in b77d79da — thanks, they were real regressions this PR introduced.

**1. Attach before `SessionInfo` resolves.** Confirmed: `useSessions()` starts at `[]` and fills only after the async `listSessions()` round-trip, while `use-host-client` has already put the connection in `attached` scope. The replay would not cover the gap — `engine-session-attach.test.ts` shows it carries status / open asks / catalogs / capabilities, not message chunks, and an in-flight claude reply is not in `history.read` yet (`conversation-store.ts` notes claude writes the row only when the message completes).

Fixed by announcing on the validated route id instead of the resolved `SessionInfo`: `useSeededConversation(sessionId, session)`. That also binds `useConversation` to the route id immediately, so the store starts consuming live events without waiting for the list. Matches the neighbouring `useSessionActions(parsed.data, …)` call.

**2. Detach through a disposed connection.** Confirmed: `SocketIoTransport.sendBytes` throws `'socket not connected'`, `send()` has no guard, and `ConnectionController.releaseGeneration` calls `client.dispose()` (line 380) synchronously *after* `onPromote(null)` (line 378) — so React processes the unmount only once the client is already dead.

Fixed at the client rather than with a try/catch in the hook, because the frame is also pointless: `Hub.removeConnection` deletes the connection's entire subscription on close. `attachSession`/`detachSession` now no-op once state is `closed`/`disposed`. Guarded on those two states only, so nothing changes for a live connection.

Both fixes carry tests, and both were mutation-checked — reverting either guard fails its test (the second with `LocalTransport: send before connect()`). Live daemon acceptance re-run and still passing: scoped client 0 events for the unattached thread, 5 replayed on attach.

PR #338, opened 2026-07-30
feat(pty): close terminal tabs on process exit

AprilNEA:
## Summary

- close right and bottom terminal tabs when their PTY process exits
- preserve exit delivery across terminal panel handoffs
- support `Cmd+W` / `Ctrl+W` for the focused terminal tab without closing the Electron window

## Verification

- `devenv shell -- pnpm check:ci`
- `devenv shell -- pnpm test`
- verified normal `exit`, forced PTY termination, and `Cmd+W` in the real Electron app

Linear: https://linear.app/arcbox/issue/CODE-349/featuipty-终端进程被杀后自动关闭对应-tab

### DIFF
--- a/apps/desktop/src/renderer/src/shell/use-desktop-shell-shortcuts.ts
+++ b/apps/desktop/src/renderer/src/shell/use-desktop-shell-shortcuts.ts
@@ -3,6 +3,7 @@ import type { PanelSide } from '@linkcode/ui/shell/panels';
 import type { WorkbenchShellNavigation } from '@linkcode/workbench';
 
 const TOGGLE_SIDEBAR_SHORTCUT = { code: 'KeyB', modifiers: ['primary'] } as const;
+const CLOSE_TERMINAL_TAB_SHORTCUT = { code: 'KeyW', modifiers: ['primary'] } as const;
 const TOGGLE_BOTTOM_PANEL_SHORTCUT = { code: 'KeyJ', modifiers: ['primary'] } as const;
 const TOGGLE_RIGHT_PANEL_SHORTCUT = {
   code: 'KeyB',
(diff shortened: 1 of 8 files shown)


apps/desktop/src/renderer/src/shell/use-desktop-shell-shortcuts.ts
```diff
@@ -20,16 +21,38 @@ const GO_FORWARD_SHORTCUT = {
...
 
 export function useDesktopShellShortcuts({
   navigation,
   owner,
+  closeBottomTerminalTab,
+  closeRightTerminalTab,
   togglePanel,
   updateSidebarOpen,
 }: UseDesktopShellShortcutsOptions): void {
+  useKeyboardShortcut({
```
lucas77778, line 40: **Blocking (macOS):** this renderer shortcut cannot reliably override the app menu's native `role: 'close'` accelerator. `apps/desktop/src/main/menu.ts` still installs `{ role: 'close' }`, which maps to Cmd+W on macOS; Electron processes that menu accelerator before dispatching the page's DOM `keydown`, so this handler may never run and the BrowserWindow closes instead of the focused terminal tab. Please intercept/route Cmd+W at the main-process input boundary (then let the renderer choose terminal tab vs window), or remove the native accelerator collision and preserve window-close behavior explicitly for non-terminal focus. Add coverage for the menu/renderer interaction; a renderer-only shortcut test would not catch this. Electron's shortcut docs describe menu accelerators as native local shortcuts, and `before-input-event` as the interception point before renderer key events: https://www.electronjs.org/docs/latest/tutorial/keyboard-shortcuts

lucas77778 (changes requested):
The PTY-exit/handoff path looks sound, but the Cmd+W path still conflicts with the native macOS Close Window accelerator. Please resolve that main-process/menu ownership conflict before merging.

PR #339, opened 2026-07-30
feat(resources): add task resources panel

AprilNEA:
## Summary

- add session-scoped Sources and Outputs resources across the wire protocol, client SDK, engine, and daemon SQLite persistence
- persist daemon-managed source uploads and inject ready source paths into ordinary agent prompts without exposing the injected block in live chat or restored provider history
- auto-register durable added deliverables and stable resource links, with regular-file validation, deduplication, safe hosting, and unlink semantics
- add the shared `TaskResourcesPanel` as an independent task-context surface: a dedicated top-right toggle with a floating card when at least 1280px is available, and an anchored Popover on narrower Desktop/Webview layouts
- hide Resources when no Thread is active; when the desktop Right Panel is open, suppress the detached floating card and anchor the Resources Popover to its toolbar button instead
- keep Resources completely separate from the existing Diff / Terminal / Browser / Files Right Panel state and layout
- configure the official Linear MCP idempotently during orb setup when `LINEAR_API_KEY` is available

Linear: [CODE-480](https://linear.app/arcbox/issue/CODE-480/featresources-add-the-task-resources-panel)

## Verification

- `pnpm format:check`
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm lint:ci` — full repository, 0 errors
- `pnpm typecheck`
- `pnpm test` — 267 files and 2120 tests passed; 1 test skipped by the existing suite
- `pnpm -F @linkcode/daemon build`
- exercised the real Desktop flow in the orb; verified the independent floating card at 1296×828 and the toolbar-anchored Popover at 1000×760 and while the desktop Right Panel is open
- verified the Popover has no modal backdrop, remains aligned to the Resources trigger, and has no clipping at either tested viewport
- verified no Resources toggle on the new-task/no-conversation surface and no Resources mutation of Diff / Terminal / Browser / Files state
- verified malformed persisted resource visibility state is discarded and `resources` is rejected as a desktop Right Panel section

`pnpm check:ci` itself could not complete in the orb because its concurrent ESLint workers retained the default 2 GiB heap and OOMed. Its three component gates were run separately; the full repository lint passed in single-process CI mode with an 8 GiB heap.

## Checklist

- [ ] `pnpm check:ci` and `pnpm test` both pass (plus `cargo fmt` / `clippy` / `test` for Rust changes)
- [x] The affected surface was run and observed working
- [x] If a wire message changed: `WIRE_PROTOCOL_VERSION` is bumped to 63 to disambiguate the parallel schema changes
- [x] New code and assets are original work, or their origin and license compatibility are noted above


AprilNEA:
image

PR #344, opened 2026-07-30
feat(plugins): plugins, MCP, and skills management on the provider-plugin model

lucas77778:
Builds the client-facing management surface for provider plugins, user-owned custom MCP servers, and skills on top of the provider-plugin model that landed with CODE-23/403/432 (which was backend-only: no wire contract, no UI, `managementCapabilities` hardcoded all-false).

Supersedes the closed #261 / CODE-382~387 batch — that design invented a parallel "capability unit + connector" concept in the same paths. Design doc: [Plugins / MCP / Skills Management](https://linear.app/arcbox/document/plugins-mcp-skills-management-design-bdf5c992be29).

## What ships

**Wire (protocol 67; minimum compatible 67)**
- `plugin.list.get`/`result` carrying `Plugin[]`, standalone skills, and `providerStatus[]` — so the UI can tell "this agent has no plugins" from "its CLI failed".
- `plugin.set-enabled`/`plugin.updated` and `skill.set-enabled`/`skill.updated`, both replying with the re-read entity so clients patch one cache entry instead of re-running discovery.
- `config.get`/`config.set` gain custom MCP servers: masked reads (`envKeys`/`headerKeys`, never values) and per-key patch ops.
- `session.started` gains optional `mcpWarnings`.

**Provider management, honestly capability-gated**
- Claude plugin enable/disable via `claude plugin enable|disable -s <scope>`; Codex enable/disable via `config/value/write` on `plugins."<id>".enabled`.
- Codex plugin install/uninstall via app-server `plugin/install` and `plugin/uninstall`; Claude installation remains unimplemented by this adapter.
- Per-skill toggles: Claude writes `skillOverrides` in settings.json (what the TUI `/skills` dialog writes; read-modify-write preserving every other key and any finer `name-only`/`user-invocable-only` tier the user set), Codex calls `skills/config/write`.
- Both providers **blind-write** — a nonexistent plugin/skill still reports success — so every write is followed by a re-read, and that readback is the only success check. Covered by tests.

**Custom (BYO) MCP servers**
- Persisted in `~/.linkcode/config.json` (0600, per-entry tolerant parse), injected into MCP-capable sessions' `StartOptions.mcpServers` alongside the existing simulator endpoint, with `agent-unsupported`/`name-conflict` warnings instead of silent drops.
- Writes are patch ops with per-key secret semantics (blank = keep, typed = replace, removed = delete). A whole-array replace would clobber real secrets with mask placeholders, so it is structurally impossible here. `buildCustomMcpPatch` is the highest-value unit test in the batch.

**Settings page** — one "Plugins & Skills" category on desktop and webview, four tabs: **Plugins** (installed only, provider-grouped, capability-gated switches), **Market** (uninstalled marketplace listings), **MCP** (custom servers + read-only plugin-provided ones), **Skills** (plugin + standalone, per-skill switches). Manual refresh only — discovery is a real CLI shell-out.

## Verification

- `pnpm check:ci` + `pnpm test` green on the current head (2341 tests).
- Dev-mock Playwright pass, 17/17: card rendering, toggle round-trip, secret never reaching the DOM, blank-save preserving stored secrets, add/edit/remove, empty states.
- Live against a real daemon before the latest rebase (then wire 64): 469 real plugins + 70 skills discovered; `claude plugin disable/enable` round-trips on the real CLI; a real skill toggle writes `skillOverrides` and restoring it removes the key; `~/.claude/settings.json` diffed byte-identical to a pre-test backup afterwards.

Two real bugs were found by that live pass and are fixed here rather than left for later:
- **Discovery was broken for both providers on any current install** (CODE-505). claude's `available[].source` became a union (bare path string in 52 of 275 entries, object with its own `source` discriminator in the rest) and its `version` is `null` for 262 of 275; codex omits `version` entirely while the schema required the key present. One bad entry failed the whole array, so the page was empty everywhere. Regression fixtures now cover every observed shape.
- Two secret-row inputs inside one base-ui `Field` both rendered `name="secrets.0.value"` (Field owns a single control), silently dropping the key field from the form.

## Deliberately out of scope

Managed/HQ connectors and OAuth (CODE-94/96/340 untouched); plugin update and Claude install/uninstall (Codex install/uninstall ships here); Claude's `name-only`/`user-invocable-only` skill tiers have no wire representation yet (preserved, not coarsened); retrofitting the same masking onto `accounts`, which still round-trips plaintext.

The Market list renders a bounded 60 entries per provider with a visible "showing 60 of 273 — narrow it down with search" note rather than truncating silently; virtualizing it is a follow-up if browsing the full catalog matters.

Closes CODE-487, CODE-488, CODE-490, CODE-491, CODE-492, CODE-493, CODE-494, CODE-495, CODE-496, CODE-497, CODE-502, CODE-503, CODE-504, CODE-505.


apps/daemon/src/config.ts
```diff
@@ -245,9 +277,14 @@ export function saveAccounts(vault: SecretVault, accounts: Accounts): void {
   writeConfigField('accounts', detachAccountSecrets(accountSecrets(vault), accounts));
 }
 
+/** Persist custom MCP servers, preserving other fields; `0600` (env/headers may hold secrets). */
+export function saveCustomMcpServers(servers: CustomMcpServer[]): void {
+  writeConfigField('customMcpServers', servers);
```
Zerlight: **[P1] Store custom MCP credentials in the vault**

This writes each complete `CustomMcpServer`, including stdio `env` values and HTTP `headers`, directly to `config.json`. These fields commonly contain long-lived credentials but bypass the existing encrypted/keyring-backed secret vault. Mode `0600` does not protect backups or Windows storage semantics, and a pre-existing loose file is written before the later `chmod`. Please add an MCP vault namespace and persist only structure plus secret references here.

packages/host/agent-adapter/src/plugins/claude-code.ts
```diff
@@ -103,24 +164,211 @@ export class ClaudeCodePluginAdapter implements PluginProviderAdapter {
...
+      if (!isRecord(document.skillOverrides)) continue;
+      for (const [name, tier] of Object.entries(document.skillOverrides)) {
+        if (typeof tier === 'string') merged[name] = tier;
+      }
+    }
+    return merged;
+  }
+
+  private settingsFileFor(scope: StandaloneSkill['scope'], cwd: string | undefined): string {
+    if (scope === 'project' && cwd) return join(cwd, '.claude', 'settings.json');
```
Zerlight: **[P1] Write project skill overrides to `settings.local.json`**

For a project skill this targets the shared `.claude/settings.json`, although Claude's `/skills` UI saves personal visibility choices to `.claude/settings.local.json` ([official documentation](https://code.claude.com/docs/en/slash-commands#override-skill-visibility-from-settings)); this adapter also reads the local file after the project file. With an opposite local override, the mutation dirties the shared config and then fails readback because the local value still wins. Without one, a personal UI action still modifies potentially tracked team configuration.

packages/host/agent-adapter/src/plugins/claude-code.ts
```diff
@@ -103,24 +164,211 @@ export class ClaudeCodePluginAdapter implements PluginProviderAdapter {
...
+  } finally {
+    if (settingsWrites.get(file) === current) settingsWrites.delete(file);
+  }
+}
+
+async function writeJsonRecordAtomic(file: string, value: Record<string, unknown>): Promise<void> {
+  const temporaryFile = `${file}.${process.pid}.${randomUUID()}.tmp`;
+  try {
+    await mkdir(dirname(file), { recursive: true });
+    await writeFile(temporaryFile, `${JSON.stringify(value, null, 2)}\n`, 'utf8');
```
Zerlight: **[P1] Preserve settings-file permissions during atomic writes**

This temporary file is created with Node's default `0666` mode masked by the process umask (`0644` under the common `022` setting). The following `rename` replaces the original inode, so toggling any skill can downgrade an existing `0600` `~/.claude/settings.json` and expose settings, environment values, or hooks to other local users. Create the temporary file with the original mode, or at least `0600`, before renaming it.

packages/foundation/schema/src/model/custom-mcp.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * A LinkCode-owned ("bring your own") MCP server, independent of any provider plugin. It is
+ * persisted in the daemon config and injected into MCP-capable sessions' StartOptions at start —
+ * never written into a provider's own config file (codex has its own `codex mcp` store).
+ */
+export const CustomMcpServerSchema = z.object({
+  /** Client-minted, stable across edits (mirrors the Account.id precedent). */
+  id: z.string().min(1),
+  enabled: z.boolean(),
+  server: McpServerSchema,
```
Zerlight: **[P2] Reject blank custom MCP names before persistence**

`McpServerSchema` accepts an empty name, and the form's `min(1)` check runs before trimming, so a whitespace-only name is persisted as an empty string. Starting an unsupported agent then creates a warning with an empty `serverName`, but `McpWarningSchema` rejects it; the transport drops the entire `session.started` frame, leaving the start promise pending after the daemon has already created a live session. Please enforce a trimmed non-empty name at both the persistence and form boundaries.

packages/host/engine/src/agent/request-handler.ts
```diff
@@ -119,6 +123,11 @@ export class AgentRequestHandler {
                   this.providers.setAccounts(accounts),
                 ),
           ).pipe(
+            Effect.andThen(
```
Zerlight, line 122: **[P2] Validate the whole `config.set` request before committing fields**

When one payload contains provider/account changes and a custom MCP patch, the earlier fields are persisted before `applyPatch` performs its conflict checks. A duplicate ID, duplicate name, reserved name, or storage failure therefore returns `request.failed` while the provider/account changes remain committed. Build and validate one next-state snapshot before writing, or make each wire request modify only one configuration resource.

packages/host/agent-adapter/src/plugins/codex.ts
```diff
@@ -86,43 +100,191 @@ const CodexPluginDetailSchema = z.object({
...
+        mergeStrategy: 'upsert',
+      });
+    });
+  }
+
+  async installPlugin(
+    id: string,
+    opts: PluginDiscoveryOptions = {},
+  ): Promise<PluginInstallOutcome> {
+    return this.withDiscoveryServer(async (server) => {
```
Zerlight: **[P2] Do not apply the discovery timeout to installs**

`installPlugin` (and `uninstallPlugin`) runs inside `withDiscoveryServer`, whose unconditional 30-second timer aborts and closes the app server. A valid network, Git, or npm-backed installation can exceed 30 seconds, causing the request to fail and potentially leaving partially applied provider state. Scope this deadline to discovery, or give mutations their own cancellation and readback lifecycle.

packages/presentation/ui/src/shell/plugins/plugins-tab.tsx
```diff
@@ -0,0 +1,132 @@
...
+                {label}
+                {market && group.plugins.length > 0 ? (
+                  <span className="font-normal text-2xs text-label-tertiary tabular-nums">
+                    {t('marketCount', { count: group.plugins.length })}
+                  </span>
+                ) : null}
+              </span>
+            }
+          >
+            {group.discoveryFailed ? (
```
Zerlight: **[P2] Keep successful plugin data visible on partial discovery failure**

The engine catches plugin and standalone-skill discovery separately and preserves the successful side's values, but either failure marks the single provider status as failed. This branch checks that status first, so a `skills/list` failure replaces a non-empty, successfully discovered plugin catalog with an error row. Render the cards alongside a facet-specific warning, or carry separate statuses for plugins and skills.

packages/client/workbench/src/settings/plugins/view.ts
```diff
@@ -0,0 +1,166 @@
...
+    key: `${plugin.provider}:${plugin.id}`,
+    provider: plugin.provider,
+    id: plugin.id,
+    title,
+    description: plugin.description,
+    version: plugin.installations.find((entry) => entry.version)?.version ?? plugin.version,
+    marketplaceLabel: plugin.marketplace?.displayName ?? plugin.marketplace?.name,
+    availability: plugin.availability,
+    installed: plugin.installations.length > 0,
+    canInstall: plugin.managementCapabilities.install,
```
Zerlight: **[P2] Gate installation on plugin availability**

`canInstall` mirrors only the management capability and ignores `availability`, despite the schema contract requiring clients to combine them. Codex maps `DISABLED_BY_ADMIN` to `blocked` while retaining `install: true`, so the Market tab exposes an actionable Install button that policy forbids. Combine the capability with availability here, and apply the same rule to enable/toggle affordances.

packages/client/workbench/src/settings/plugins/plugins-settings.tsx
```diff
@@ -0,0 +1,189 @@
...
+            ),
+          },
+        { revalidate: false },
+      );
+      return;
+    }
+    const separator = row.pluginKey.indexOf(':');
+    const provider = row.pluginKey.slice(0, separator);
+    if (provider !== 'claude-code' && provider !== 'codex') return;
+    const updated = await toggle.trigger({
```
Zerlight: **[P2] Do not drive per-skill state with a plugin-level toggle**

A bundled-skill row is checked from the component's `skill.enabled`, but this handler calls the whole plugin's `setPluginEnabled`. For the valid state where a Codex plugin is enabled while one bundled skill is disabled, clicking the switch merely re-enables the already enabled plugin; readback leaves the skill disabled and the switch never changes. Either represent plugin enablement in this row or expose a real per-skill mutation.

packages/client/workbench/src/settings/plugins/mcp-settings.tsx
```diff
@@ -0,0 +1,332 @@
...
+  onClose,
+  onSubmit,
+}: {
+  previous: CustomMcpServerPublic | undefined;
+  busy: boolean;
+  onClose: () => void;
+  onSubmit: (draft: CustomMcpServerDraft) => void;
+}): React.ReactNode {
+  const t = useTranslations('settings.plugins.mcp');
+  const { control, register, handleSubmit } = useForm<McpForm>({
```
Zerlight: **[P2] Surface MCP form validation errors**

The form installs a resolver but never reads `formState.errors` or field state, and no `FieldError` is rendered. An empty name, command, or URL therefore makes `handleSubmit` silently refuse to submit while the dialog stays open with no feedback. Bind each validation error to its field and expose an accessible message.

Zerlight (changes requested):
I found three high-priority safety issues and seven correctness or UX issues. The most serious problems are that custom MCP credentials bypass the existing secret vault, while Claude skill toggles target the wrong settings layer and can weaken settings-file permissions. Please address the inline comments before merging.

Verification: 136 focused tests passed. In the full suite, 2,312 tests passed and one unrelated timeout passed immediately when rerun in isolation. The wire-version change from 66/66 to 67/67 is consistent with the new required `config.get.result.customMcpServers` field.

apps/daemon/src/config.ts
lucas77778: Fixed in 16022d7c. Custom MCP env/header values now live in a dedicated `custom-mcp` SecretVault namespace; `config.json` stores only structure plus key/null placeholders. This also lazily migrates legacy inline credentials, uses collision-safe encoded refs, and rolls back vault state if either vault or config persistence fails. Added daemon config/vault coverage.

packages/host/agent-adapter/src/plugins/claude-code.ts
lucas77778: Fixed in 16022d7c. Project-scoped Claude skill visibility now writes `.claude/settings.local.json`, so personal overrides no longer dirty the shared project settings file. Readback follows the same effective precedence, including an explicit local `on` when shared settings are `off`.

packages/host/agent-adapter/src/plugins/claude-code.ts
lucas77778: Fixed in 16022d7c. Atomic Claude settings writes now preserve the existing file mode; newly created settings files default to `0600`. Tests cover mode preservation and the secure default.

packages/foundation/schema/src/model/custom-mcp.ts
lucas77778: Fixed in 16022d7c. Custom MCP names are trimmed and required to be non-empty in stored/public/update schemas and at the form boundary, without tightening the generic `McpServerSchema`. Added schema and UI validation coverage.

packages/host/engine/src/agent/request-handler.ts
lucas77778: Fixed in 16022d7c by constraining `config.set` to mutate exactly one of providers, accounts, or custom MCP servers per request. Cross-resource payloads now fail schema validation before any persistence, preventing partial commits and stale snapshot overwrites. Added wire contract tests.

packages/host/agent-adapter/src/plugins/codex.ts
lucas77778: Fixed in 16022d7c. Codex install/uninstall now use a mutation lifecycle without the 30-second discovery timer; discovery/list/toggle/skill operations retain the existing timeout. Added long-running mutation coverage.

packages/presentation/ui/src/shell/plugins/plugins-tab.tsx
lucas77778: Fixed in 16022d7c. The Plugins tab now renders successfully discovered plugin cards even when the provider reports a partial discovery failure, with the failure shown alongside the retained data. Added view/render coverage.

packages/client/workbench/src/settings/plugins/view.ts
lucas77778: Fixed in 16022d7c. Install and enable affordances now require `availability === "available"` in addition to the corresponding capability. Uninstall remains capability-based so blocked/unavailable installed plugins can still be removed. Added view-model tests.

packages/client/workbench/src/settings/plugins/plugins-settings.tsx
lucas77778: Fixed in 16022d7c. Bundled skills no longer invoke the plugin-level toggle; they are rendered read-only with their enabled/disabled state. Standalone skills retain their actual per-skill toggle. Added a focused Skills tab regression test.

packages/client/workbench/src/settings/plugins/mcp-settings.tsx
lucas77778: Fixed in 16022d7c. The MCP editor now uses semantic form submission, trims the name, renders accessible `FieldError` messages for name/command/URL validation, and gives non-submit controls explicit button types.

apps/daemon/src/config.ts
```diff
@@ -245,10 +283,53 @@ export function saveAccounts(vault: SecretVault, accounts: Accounts): void {
...
 }
 
+/** Persist custom MCP structure and key names; values go to the vault. */
+export function saveCustomMcpServers(
+  vault: SecretVault,
+  servers: CustomMcpServer[],
+  previous: CustomMcpServer[],
+): void {
+  const store = customMcpSecrets(vault);
+  const stripped = detachCustomMcpSecrets(store, servers);
```
Zerlight: [P2] Keep the vault and config crash-consistent

`detachCustomMcpSecrets` durably replaces the `custom-mcp` vault namespace before `config.json` is replaced. The catch handles an ordinary write error, but not process or host termination between the two renames. After removing or renaming a secret key, the old config then references a missing vault key; the next load leaves that placeholder as `null`, schema validation fails, and the entire server is dropped. Please use versioned or journaled state so startup can recover one complete snapshot.

packages/host/agent-adapter/src/plugins/codex.ts
```diff
@@ -86,103 +100,317 @@ const CodexPluginDetailSchema = z.object({
...
+        mergeStrategy: 'upsert',
+      });
+    });
+  }
+
+  async installPlugin(
+    id: string,
+    opts: PluginDiscoveryOptions = {},
+  ): Promise<PluginInstallOutcome> {
+    return this.withServer(async (server) => {
```
Zerlight: [P2] Make long-running mutations cancelable

Allowing install and uninstall to exceed the discovery deadline is correct, but this path now has neither a timeout nor caller cancellation. The Engine wraps these mutations in zero-argument `Effect.tryPromise` callbacks, so interruption does not reach this internal `AbortController`; a stuck handshake or RPC can therefore leave the Codex app-server alive and prevent its `finally` cleanup from running. Please propagate the Engine abort signal through the adapter lifecycle, while still allowing successful operations to run beyond 30 seconds.

packages/host/agent-adapter/src/plugins/claude-code.ts
```diff
@@ -103,24 +164,224 @@ export class ClaudeCodePluginAdapter implements PluginProviderAdapter {
...
+  const temporaryFile = `${file}.${process.pid}.${randomUUID()}.tmp`;
+  try {
+    let mode = 0o600;
+    try {
+      mode = (await stat(file)).mode & 0o777;
+    } catch (error) {
+      if (!isErrorLikeObject(error) || !('code' in error) || error.code !== 'ENOENT') throw error;
+    }
+    await mkdir(dirname(file), { recursive: true });
+    await writeFile(temporaryFile, `${JSON.stringify(value, null, 2)}\n`, {
```
Zerlight, line 277: [P2] Restore the exact destination mode

Creating the temporary file with `{ mode }` still applies the process umask. For example, an existing `0660` settings file becomes `0640` under a typical `022` umask, so this does not preserve permissions exactly. Please apply `chmod` to the temporary file before renaming it and cover a mode affected by the umask in the test.

Zerlight (changes requested):
Thanks for addressing the first review. The earlier UI, schema, and persistence issues are largely resolved, and the focused verification is green. Three correctness issues remain: custom MCP persistence is not crash-consistent across the vault/config pair, Codex install and uninstall have no shutdown cancellation path, and the Claude settings rewrite still narrows some existing file modes through the process umask. Please address the inline comments before merging.

packages/host/agent-adapter/src/plugins/claude-code.ts
lucas77778: lucas77778 addressed this in e2a3f5ae. The atomic JSON rewrite now explicitly chmods the temporary file after creation and before rename, so the destination mode is restored independently of umask. The regression test preserves 0660 while running under umask 022.

packages/host/agent-adapter/src/plugins/codex.ts
lucas77778: lucas77778 addressed this in e2a3f5ae. Engine mutations now use the AbortSignal supplied by Effect.tryPromise and pass it through PluginDiscoveryOptions; the Codex lifecycle forwards that signal into its internal controller while keeping the 30-second deadline scoped to discovery only. Install and uninstall cancellation are covered at both the adapter and Engine fiber-interruption layers, including cancellation during app-server startup.

apps/daemon/src/config.ts
lucas77778: lucas77778 addressed this in e2a3f5ae. Custom MCP snapshots now carry a generation, secret refs are generation-qualified, and the config rename is the commit point: the vault is staged with both the previous and next generations first, then old refs are pruned best-effort after the config commit. A pre-commit write failure restores the prior vault snapshot, and legacy snapshots/refs remain readable. Regression coverage now exercises interruption before the config rename, interruption before pruning, rollback/prune failures, and legacy migration.

lucas77778:
Addressed all findings from Copilot’s two completed reviews:

- `2f84c5e7`: corrected stale comments/copy, ICU component plurals, and the PR wire-version description.
- `2bdf2100`: restored the singular `componentKind` locale keys used by settings search while retaining plural-aware counts.
- `9320b3e8`: preserves an absent `mcpServers` value when every custom server is skipped, and renders a skeleton rather than the empty state while custom MCP data is loading. Added regression coverage for both semantics.

Validation: `pnpm check:ci` (the final lint pass used the equivalent serial `pnpm lint:ci` after an auto-concurrency worker OOM), `pnpm test` (2342 passed), and the webview production/mock browser smoke all pass.

The latest requested Copilot re-review did not run because the requester had reached the Copilot review quota.

AprilNEA (approved):

PR #350, opened 2026-07-30
feat(desktop,agent-adapter): API-key / relay login on the signed-out agent card

lucas77778:
Closes CODE-344.

## What

The signed-out agent card only offered the CLI's OAuth login. It now also offers **"Use an API key"**, which opens a dialog for a relay/gateway key:

- **Name / Base URL / credential type (`x-api-key` vs `Bearer`) / secret / default model**, with the protocol pre-set to the agent's native one (claude-code → Anthropic, codex → OpenAI Chat).
- **"Detect models"** asks the daemon to read the endpoint's own model list; the model field turns into a picker. A rejection (bad key, unreachable host) is shown verbatim, including the vendor's own reason.
- Saving stores it as a normal **pool account** (CODE-127 contract) and binds it as that agent's `activeAccountId` — the credential/base-URL injection at session start already existed for every adapter (`ANTHROPIC_BASE_URL`, `OPENAI_BASE_URL`, opencode `provider.options`, pi runtime key).
- The card is offered on both the idle and the failed-login state — a failed OAuth attempt is exactly when the API-key path is wanted.

**Bug fixed along the way:** the login cue was only suppressed by the *legacy* `providers[kind].apiKey`, so a bound key account left the "needs sign-in" card up even though the daemon injects the key at spawn. Now any bound account carrying a key/token clears it; an `oauth` account (no secret — it delegates back to the CLI store) still does not.

## New wire op

`config.probe-models` (+ `.result`) — the client cannot read a relay's model list itself: the renderer's CSP blocks remote fetches, and only the daemon holds the secret. Existing paths could not serve it either: claude-code's `startCatalog` returns a static empty list, opencode's `provider.list` ignores the injected credential, and `agent.catalog` reads only *saved* config.

Addressing follows each protocol's base-URL convention — Anthropic-shaped URLs are bare origins (`{base}/v1/models`, `x-api-key` + `anthropic-version`), OpenAI-shaped ones already end in `/v1` (`{base}/models`, `Bearer`). Bare-array responses (common in relays) are accepted alongside the vendors' `{data}` envelope.

**`WIRE_PROTOCOL_VERSION` 63 → 64** (Invariant 1): rebuild and restart the daemon together with every client.

## Verification

- `pnpm check:ci` + `pnpm test` (pre-existing, unrelated failure: `worktree-service.test.ts` temp-path comparison on macOS, fixed on another branch).
- New end-to-end engine test drives `config.probe-models` through the real router/handler against a local HTTP relay — both the model list and the 401-with-reason path.
- Unit tests for URL/header/parse shapes, the cue suppression (key account vs oauth vs stale binding), and the card's new button.
- **Not observed in the running desktop app**: reproducing the signed-out card needs the local claude/codex CLI signed out. The renderer half is covered by tests only — worth a manual pass before merge.

## Scope note

Only claude-code and codex have an auth probe, so they are the kinds that show a signed-out card today; opencode/pi/grok-build surface no such card, and the entry appears automatically if they gain one. Relay accounts for those agents are still reachable through Settings → Providers.


lucas77778 (commented):
Review findings — five issues need fixing before merge (GitHub does not allow the PR author to submit a Request changes review on their own PR):

1. High: `AgentApiKeyLoginDialog` allows Save before accounts/providers load; `accounts ?? []` and `providers ?? {}` can replace existing daemon configuration with incomplete snapshots.
2. High: `config.probe-models` accepts arbitrary URLs and follows redirects, enabling daemon-side SSRF and potential cross-origin `x-api-key` disclosure. Restrict schemes/destinations, reject redirects, and bound successful responses.
3. Medium: account creation and provider binding are two non-atomic, non-idempotent writes; partial failure leaves orphan/duplicate accounts. Use one daemon-owned atomic create-and-bind operation.
4. Medium: detected models remain selectable after Base URL, protocol, credential type, or secret changes, including in-flight races. Key results to the probe inputs and discard stale responses.
5. Low: `providers[kind]?.apiKey !== undefined` treats an empty legacy key as a usable injected credential and incorrectly hides the login card. Require a non-empty trimmed key.

lucas77778:
[CODE-344-1.webm](https://uploads.linear.app/51848c28-516b-472f-8ce1-b64be2222039/7cf0e1d7-7424-44b2-bcd7-c1145fa4154c/badad66f-cbc1-4855-8502-793a4fa198ac?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzUxODQ4YzI4LTUxNmItNDcyZi04Y2UxLWI2NGJlMjIyMjAzOS83Y2YwZTFkNy03NDI0LTQ0YjItYmNkNy1jMTE0NWZhNDE1NGMvYmFkYWQ2NmYtY2JjMS00ODU1LTg1MDItNzkzYTRmYTE5OGFjIiwiaWF0IjoxNzg1NDc4Mjc0LCJleHAiOjE4MTcwNDg4MzR9.bWm_UjtsrYOCS44tG3jubVwb7TCTF3XeL1Vo27ruXys)

[CODE-344-2.webm](https://uploads.linear.app/51848c28-516b-472f-8ce1-b64be2222039/8a15024a-3719-4bed-bbd9-bba7b9da7964/82033a39-60f2-47ee-af17-2c4123396e27?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzUxODQ4YzI4LTUxNmItNDcyZi04Y2UxLWI2NGJlMjIyMjAzOS84YTE1MDI0YS0zNzE5LTRiZWQtYmJkOS1iYmE3YjlkYTc5NjQvODIwMzNhMzktNjBmMi00N2VlLWFmMTctMmM0MTIzMzk2ZTI3IiwiaWF0IjoxNzg1NDc4Mjc0LCJleHAiOjE4MTcwNDg4MzR9.-Kd4WhrMgq9NX3hPIvwjXXvz4UOcw2f4UXMRE-seQo8)

AprilNEA:
不太对，这里最好是跳到 Settings 里面

lucas77778:
那还是改为go to setting，因为我想的是在这里快捷填完就不用在setting里填完再跳转回new task

Zerlight (changes requested):
Requesting changes on head `15daf721`.

1. **[P1] Require HTTPS before sending credentials.** `resolvePublicEndpoint` accepts public `http:` URLs, and `requestPublicModelList` then sends `Authorization` or `x-api-key` over plaintext HTTP. Private/loopback HTTP is already rejected, so this only enables remote cleartext credential transmission.

2. **[P1] Reject unsupported agent/protocol combinations.** The dialog offers every `AccountProtocolSchema` option for every agent, while the capability matrix says, for example, that Codex supports only OpenAI Chat/Responses. `accountBinding` persists any combination without validation. Selecting Codex + Anthropic therefore succeeds, suppresses the onboarding cue, and leaves session startup with an unusable binding. Filter the UI and validate again at the daemon boundary.

3. **[P2] Surface save failures.** The submit handler stores failures under React Hook Form's `root` error, but `rhfErrorsToFormErrors` explicitly drops `root` and the dialog renders no separate root error. A rejected valid submission silently re-enables the button without telling the user what failed.

4. **[P2] Honor the advertised `x-api-key` credential type.** The UI labels `api-key` as “API key (x-api-key)”, but every OpenAI-shaped request converts both credential variants to `Authorization: Bearer`. OpenAI-compatible relays that require `x-api-key` cannot be detected or used. Either send the selected header or remove/restrict the unsupported choice.

5. **[P2] Do not allow a save to outlive dismissal.** Cancel, Escape, and backdrop dismissal remain enabled while `save.isMutating`. Closing the dialog does not cancel or invalidate the request, so the account may still be created after an apparent cancellation; the stale completion can also call the shared `close` action after another agent's dialog has opened.

6. **[P2] Make create-and-bind atomic across the vault and `config.json`.** `saveProviderConfiguration` mutates both secret namespaces via `replaceAll` before writing the JSON file. If the file write fails, the caller sees a failure while new/orphaned secrets remain; a stable-ID credential update can take effect after restart despite the failed operation.

7. **[P3] Preserve bounded non-2xx error details.** Error responses are destroyed as soon as they exceed 200 bytes, so the later `.slice(0, ERROR_BODY_LIMIT)` path is unreachable for a 201-byte vendor response. The UI receives only “response exceeded 200 bytes” instead of the intended status and truncated vendor reason.

8. **[P3] Implement `config.probe-models` in the dev mock.** The client sends a correlated request, but the mock switch has no matching case and its default branch silently drops the frame. “Detect models” therefore remains pending forever in mock mode.

9. **[P3] Retry all validated DNS addresses.** `resolvePublicEndpoint` validates every result but returns only `addresses[0]`. A dual-stack or multi-A endpoint fails when the first address is unreachable even if another validated address works.

Additional merge blockers:

- The latest PR discussion agreed to navigate to Settings, but the current head still mounts and opens the modal.
- The PR currently conflicts with `master` in `workbench.tsx` and `wire/message.ts`. Current `master` is `WIRE_PROTOCOL_VERSION = 66` / `MIN_COMPATIBLE_WIRE_VERSION = 66`; after replaying this additive wire operation, the resolved values should be wire `67` and minimum compatible `66`.
- CI is still red. The failure appears to be an unrelated flaky daemon runtime PID/port fixture, and 98 targeted tests passed locally, but the merge gate still needs a green rerun.

The earlier self-review items covering save-before-load overwrites, redirect SSRF, stale model detection, and whitespace-only legacy keys appear fixed in this head.

AprilNEA (approved):

PR #368, opened 2026-08-01
feat(schema,agent-adapter,ui): carry each agent's own configured default model/effort in the start catalog

AprilNEA:
Closes CODE-524.

The New Task page showed a model it had guessed rather than the one the session would actually run. With no LinkCode provider config and no remembered pick, `displayedModel` fell through to the hardcoded `AGENT_DEFAULT_MODELS` table while the session started on whatever the agent CLI's own configuration said. The effort chip had the same hole with no fallback at all, so a `~/.codex/config.toml` carrying `model_reasoning_effort` rendered as an empty chip.

The pre-session catalog is the natural carrier and already reaches the surface per-workspace via `useAgentStartCatalogs`, but `AgentStartCatalogSchema` had nowhere to put it. The approval-policy axis has done this correctly for a while — claude-code's `startCatalog` resolves the tier from the user's `settings.json`. This closes the same gap on the model/effort axis.

## Precedence

Only layer 4 is new. It sits below the remembered pick deliberately: a past explicit choice outranks a config file the user did not make through LinkCode.

1. Explicit pick on the page
2. `preferredModels` — the remembered last pick
3. `providers[kind].defaultModel` / `account.model` — LinkCode settings
4. **The agent CLI's own config** ← new
5. `AGENT_DEFAULT_MODELS` hardcoded fallback

## Changes

- **schema** — `AgentStartCatalogSchema` gains optional `defaultModel` / `defaultEffort`. `WIRE_PROTOCOL_VERSION` 66 → 67; `MIN_COMPATIBLE_WIRE_VERSION` unchanged, since the change is purely additive.
- **codex** — new `codexConfiguredModel()` reads `model` / `model_reasoning_effort` with the active profile resolved, sharing one config.toml read with `codexConfiguredSandbox`. `startCatalog` prefers it over the served `model/list` default and falls back to the named model's own `defaultReasoningEffort`.
- **claude-code** — `startCatalog` resolves both axes through the SDK's `resolveSettings({cwd})`, the same call the start path already uses, so the managed and remote policy tiers a raw settings.json walk would miss are applied. A load failure leaves the axes absent rather than sinking the approval tiers, which are readable without the SDK.
- **pi** — `defaultModel` is the registry's first entry, exactly what an unpicked session falls back to.
- **opencode** — deliberately absent. `provider.list` carries a per-provider default map, but which one a fresh session adopts is unverified and no local install was available to check against.
- **grok-build** — left alone; its only model is the static table's single entry, so filling it would just re-hardcode the guess.
- **ui** — the surface consumes both at layer 4. Effort needed the display-vs-submit split the model already had: `displayedEffort` shows the catalog value, dropped when the displayed model has no such level, while only a real pick still travels to the adapter.

## Verification

Observed running — dev desktop against the real HOME with a throwaway `LINKCODE_PROFILE`, so nothing remembered and no provider config could shadow the catalog:

| agent | its own config | composer chip |
| --- | --- | --- |
| claude-code | `effortLevel: "xhigh"`, no `model` | `Sonnet 5 · xHigh` |
| codex | `model = "gpt-5.6-sol"`, `model_reasoning_effort = "ultra"` | `GPT-5.6-Sol · Ultra` |

Both effort values are new; that chip was empty before. claude-code's model correctly stays on the built-in `Sonnet 5` because its settings declare none. The codex model is weak evidence alone — `gpt-5.6-sol` happens to equal the hardcoded fallback — but it travels the same struct, wire field, and UI line as the effort, and the config-outranks-served-default precedence is covered by unit tests.

Also: `codexConfiguredModel` against a real `~/.codex/config.toml`, 7 cases over a throwaway `CODEX_HOME`, 3 codex `startCatalog` precedence cases, and 4 surface cases (shown-not-submitted, effort dropped for an incompatible model, LinkCode config outranks, remembered pick outranks). Whole repo: `check:ci` clean, `pnpm test` 2266 passed.

One fix worth calling out: the existing codex catalog test was reading the developer's own `~/.codex` through an unstubbed seam, so its assertions were machine-dependent. `TestCodex` now stubs `readConfiguredModel` the way it already stubbed the sandbox.

## Out of scope

opencode's axis stays open — it needs a live server to determine which provider default a fresh session adopts. claude-code's path has no test of its own; `resolveSettings` is `@alpha` in SDK 0.3.206.

### DIFF
--- a/packages/foundation/schema/src/model/agent/input.ts
+++ b/packages/foundation/schema/src/model/agent/input.ts
@@ -110,6 +110,11 @@ export const AgentStartCatalogSchema = z.object({
   models: z.array(AgentModelOptionSchema),
   policies: z.array(ApprovalPolicySchema),
   defaultPolicyId: ApprovalPolicyIdSchema.optional(),
+  /** What the agent's OWN configuration would start on, read before any session exists — display
+   * only, on the same terms as `defaultPolicyId`: sending either back would read as an explicit
+   * pick and override the resolution the agent does for itself. Absent when unreadable. */
+  defaultModel: z.string().min(1).optional(),
+  defaultEffort: EffortLevelSchema.optional(),
 });
 export type AgentStartCatalog = z.infer<typeof AgentStartCatalogSchema>;
 
--- a/packages/foundation/schema/src/wire/message.ts
+++ b/packages/foundation/schema/src/wire/message.ts
@@ -9,7 +9,7 @@ import { WIRE_PAYLOAD_KINDS, WirePayloadSchema } from './payload';
  */
 
 /** Stamped on every frame this build sends; bump on any wire schema change. */
-export const WIRE_PROTOCOL_VERSION = 66 as const;
+export const WIRE_PROTOCOL_VERSION = 67 as const;
 
 /** The oldest `v` this build still accepts. Bump only for a breaking change — a variant or field
  * removed, renamed, or given a new meaning; additive changes leave it alone. */
--- a/packages/host/agent-adapter/src/__tests__/codex-config.test.ts
+++ b/packages/host/agent-adapter/src/__tests__/codex-config.test.ts
@@ -2,7 +2,7 @@ import { mkdtemp, rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
-import { codexConfiguredSandbox } from '../native/codex';
+import { codexConfiguredModel, codexConfiguredSandbox } from '../native/codex';
 
 /** Exercises the real config.toml read + parse + profile/top-level resolution through a throwaway
  * `CODEX_HOME` — mocking the parser would test nothing. */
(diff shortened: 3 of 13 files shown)


AprilNEA:
Both P2 findings were real; fixed in `ae34c31b` and `05bd9db1`.

**Read Codex defaults from the app-server environment** — confirmed: `codexEnv` does `Object.assign(env, cred.extraEnv)`, so an account can absolutely move `CODEX_HOME`, and `openThread` already reads the sandbox from the *merged* environment. The catalog read was the odd one out. Now built once and passed to both `startAppServer` and `readConfiguredModel`.

**Recompute effort when a model override wins** — confirmed, and it is exactly the class of lie this PR exists to remove. The catalog effort is now used only when the catalog's own `defaultModel` is the one on display; otherwise the displayed model's advertised `defaultEffort` is used.

One correction to the suggested shape, caught by re-running the desktop probe: gating purely on `displayedModel === catalog.defaultModel` regresses claude-code. Its settings declare `effortLevel` but no `model`, so `defaultModel` is absent, the displayed model is the built-in fallback, and a strict equality gate would drop the user's configured effort. Claude's effort setting is model-independent, so a catalog that names no default model keeps applying its effort:

```ts
const catalogEffort =
  catalog?.defaultModel === undefined || catalog.defaultModel === displayedModel
    ? catalog?.defaultEffort
    : modelOption?.defaultEffort;
```

**Verification** — both new tests were confirmed to fail against the pre-fix code (and only those two), so they discriminate rather than restate the implementation. Re-ran the desktop probe after the fixes: claude-code still `Sonnet 5 · xHigh`, codex still `GPT-5.6-Sol · Ultra`. Whole repo: `check:ci` clean, `pnpm test` 2269 passed.

PR #408, opened 2026-08-03
fix(billing): hand off billing to hosted console

lucas77778:
## Summary

- Replace LinkCode's balance, offers, orders, subscription, portal, custom amount, and checkout implementation with a single hosted billing handoff built by the official `@linkcode/cloud` SDK.
- Desktop exposes only `openHostedBilling()`: main registers the current `linkcode://` / `linkcode-dev://` scheme, builds `https://console.linkcode.ai/billing?returnTarget=...`, and opens it in the system browser.
- Webview builds the credential-free hosted URL through the same SDK and opens it in a new browser tab.
- Reduce the shared UI contract to `BillingSettingsPanel({ onOpenBilling })`; LinkCode does not request, store, render, or infer billing/checkout/payment state.
- Remove the sidebar balance shortcut and all LinkCode billing API sources, providers, IPC channels, and preload methods.

Linear: CODE-567
Cloud counterpart: arcboxlabs/linkcodehq#39 at `5438df8c8083f349dc3b8ed3eccdd61ca9d86358`

## Contracts

- Cloud SDK: `createHostedBillingUrl(options?: { returnTarget?: string }): string`; `HostedBillingUrlOptions` is exported.
- Hosted surface: `https://console.linkcode.ai/billing[?returnTarget=<encoded>]`.
- Desktop return target: `${scheme}://billing/return`, where scheme is `linkcode` in packaged builds and `linkcode-dev` in development.
- Cloud's existing native return bridge appends authoritative `order=<id>` and, on cancellation, `canceled=true`. Desktop deliberately does not parse either parameter and does not process checkout; the existing generic deep-link path only activates/focuses the app.
- Until the initial public npm bootstrap of `@linkcode/cloud@0.1.0`, this repository consumes the official package tarball built from the Cloud commit above, pinned at SHA-256 `576bf95543503615204163609c49fe66ad38a3081babf5b93369a94842acdb7b`.

## Verification

- `NODE_OPTIONS=--max-old-space-size=6144 pnpm check:ci` — passed: format, repo-wide ESLint (0 errors; 377 existing warnings), and typecheck.
- `pnpm test` — passed: 307 files / 2443 tests; 1 existing skipped file/test.
- Hosted handoff unit tests — passed for Desktop main, Webview route, and shared Billing panel.
- `pnpm -F @linkcode/desktop run build` — passed.
- `pnpm -F @linkcode/webview run build` — passed.
- `pnpm -F @linkcode/webview e2e:browser` — passed production billing-route smoke plus bundled mock router/wire/reload smoke.
- Bundle inspection found the Cloud hosted URL and no `checkout-sessions`, billing summary/orders, or portal-session endpoints.
- Built Desktop Electron smoke — opened the Billing settings page, launched a second instance with `linkcode-dev://billing/return?order=...- Commit hookscanceled=true`, observed one restored/focused window, and observed no local checkout-result UI.
- Commit hooks — passed large-file policy, formatting/imports, lint, and typecheck.

## Checklist

- [x] No wire message changed
- [x] The vendored SDK tarball comes from the official Cloud package/commit documented above


PeronGH (commented):
Two correctness issues on the typed gateway-billing path; details inline.

packages/host/agent-adapter/src/native/claude-code.ts
```diff
@@ -1388,6 +1388,16 @@ export class ClaudeCodeAdapter extends BaseAgentAdapter {
...
+      if (
+        this.emitGatewayError({
+          statusCode: msg.api_error_status,
+          message: msg.result,
+        })
+      ) {
+        this.teardown();
+        if (!cancelling) this.emitStatus('idle');
+        return;
+      }
```
PeronGH: This can never fire, so claude-code never gets a typed billing error.

The LinkCode Gateway account is `protocol: 'openai-chat'` (`catalog.ts`), which is exactly the combination `translationUpstream()` claims: `packages/host/engine/src/agent/translator.ts:35-37` matches `kind === 'claude-code'` + `config.protocol === 'openai-chat'`, and `SessionStartOptionsResolver.resolve` (`session/start-options-resolver.ts:46-63`) then runs `withTranslatorEndpoint`, which overwrites `config.baseUrl` with the loopback sidecar URL and sets `protocol: 'anthropic'`. By the time `start()` stores `opts`, `this.opts.config.baseUrl` is `http://127.0.0.1:PORT`, so `gatewayHost()` in `linkCodeGatewayError` is always false here — this block, and the `emitProviderError` on the non-success path below, are dead for gateway sessions. Codex/opencode/pi are unaffected because `translationUpstream` returns undefined for them and their `baseUrl` stays the gateway host.

Net effect: on the flagship agent a gateway 402 surfaces as raw error text with no top-up button. The upstream URL is spread away by `withTranslatorEndpoint`, so nothing downstream can recover it — the fix belongs in the resolver (carry the pre-rewrite upstream in `config`, e.g. `upstreamBaseUrl`, and have `linkCodeGatewayError` consult it). Worth confirming first that the `arcboxlabs/aigateway` sidecar even forwards the upstream 402 status/body, otherwise the typed error can't be reconstructed on this path at all.

Separately, and regardless of how the routing is resolved: `msg.result` is the final assistant text on `subtype === 'success'`, and `typedCode` substring-matches it for `insufficient_credits`. Once host detection starts working here, a reply that merely mentions the token converts a successful turn into an error and `return`s before `emitUsage`/`emitStop`, losing the turn's usage and stop event. Only `statusCode` is trustworthy on the success path — drop `message: msg.result` from this call.

packages/presentation/i18n/src/locales/en.ts
```diff
@@ -136,6 +136,11 @@ export const en = {
       compacting: 'Compacting context…',
       compacted: 'Context compacted',
       compactedTokens: '{pre} → {post} tokens',
+      insufficientCreditsTitle: 'LinkCode credits needed',
+      insufficientCreditsHint: 'Top up your balance, then retry this message.',
+      topUpCredits: 'Top up credits',
+      billingUnavailableTitle: 'LinkCode billing is temporarily unavailable',
+      billingUnavailableHint: 'Your request was not sent to the model. Please retry shortly.',
```
PeronGH: This contradicts Cloud's documented 503 semantics. `apps/web/content/docs/reference/ai-api-errors.mdx` in `linkcodehq` says of `503 billing_unavailable`:

> This error can occur either before an upstream call **or after a provider response while the charge is being persisted**, so do not assume that no completion was generated.

The gateway backs that up: `chat-completions.ts:490` returns `billing_unavailable` on the *post-response* persistence failure, after the upstream completion already happened (and was possibly already streamed). Telling the user "Your request was not sent to the model" is therefore wrong in that case, and it invites a retry that duplicates a completion the org may already owe for.

Something like "Billing could not record this request — it may or may not have reached the model. Retry shortly." keeps the retry advice without the false guarantee.

`insufficientCreditsHint` two lines up needs no change: the same doc guarantees 402 is generated before any upstream provider call and that retrying after a top-up is safe.

PeronGH (commented):
Correction to my earlier 503 comment (retracted inline — the mapping is contract-backed), and the real problem the Cloud contract exposes instead.

PeronGH (approved):
Approving the change itself — but **do not merge until the Cloud side is merged *and deployed*.**

The gate is the deploy, not the merge. `https://console.linkcode.ai/billing` does not exist in production today: it and `/zzz-does-not-exist` both return 200 with the identical `[organizationSlug]` catch-all entry (`<title>LinkCode</title>`), i.e. prod treats `billing` as an org slug. The hosted route only exists on `feat/code-567-hosted-billing-sdk`, and linkcodehq has no deploy workflow — the console ships via a manual `pnpm --filter @linkcodehq/console deploy`. So merging arcboxlabs/linkcodehq#39 alone still leaves `openHostedBilling()` opening a dead end, and desktop cuts releases off master.

Order:

1. Land arcboxlabs/linkcodehq#39 — its only conflict is `pnpm-lock.yaml` (checked with `git merge-tree`), so resolving it cannot change `packages/cloud/src` or the route, and the pin below stays valid.
2. Deploy the console, then verify `https://console.linkcode.ai/billing?returnTarget=linkcode://billing/return` renders the Billing entry and preserves the target through sign-in + organization resolution.
3. Merge this PR.

What is **not** blocking:

- The npm publish. `packages/vendor/linkcode-cloud-0.1.0.tgz` is sha256 `576bf95543503615204163609c49fe66ad38a3081babf5b93369a94842acdb7b` as documented, and its `dist/index.js` is faithful to `packages/cloud/src/index.ts` at `5438df8c`. The `cloud-v0.1.0` tag and npm bootstrap are a follow-up.
- Everything outside the hosted handoff. The Gateway provider, the typed 402/503 path, and API-key creation depend only on Cloud work already merged and live: `gateway.linkcode.ai/v1/models` → 401, `api.linkcode.ai/auth/get-session` → 200, `/auth/api-key/create` → 400.

Both of my earlier findings are addressed — `fix(gateway): preserve translated billing errors` for the dead typed-error path on claude-code, and `billingUnavailableHint` no longer promises the request never reached the model.

Please open a follow-up issue for the vendored tarball: once `@linkcode/cloud@0.1.0` is on npm, swap the `file:` dependency for the registry package and delete the blob (`@linkcode/tunnel@0.1.0` is the precedent). That swap will need a `minimumReleaseAgeExclude` entry for the fresh publish.

PR #454, opened 2026-08-17
fix: prevent daemon fsync crash and Pi catalog failures

### LINKED ISSUE #437: [Bug] Windows: daemon crash loop with EPERM on fsync during credential migration -> cannot connect to daemon (127.0.0.1:19523)
## Summary

On Windows, LinkCode 0.20.0 enters a permanent daemon crash loop during the one-shot "move credentials out of config.json into the secret vault" migration. The desktop app then reports it cannot connect to the daemon at `http://127.0.0.1:19523`.

## Environment

- LinkCode desktop 0.20.0 (also present since the migration feature shipped)
- Windows 11 x64, NTFS, no AV beyond Windows Defender
- App installed at `%LOCALAPPDATA%\Programs\LinkCode`

## Symptom

- UI: "无法连接到 daemon (http://127.0.0.1:19523)" / cannot connect to daemon; the daemon never listens on 19523.
- `%APPDATA%\LinkCode\logs\main.log` shows the daemon exiting with code 1, restarting ~6 times, then giving up:

```
[info]  {"operation":"config.load","msg":"Moving credentials out of config.json into the secret vault"}
[info]  {"level":60,"err":{"type":"Error","message":"EPERM: operation not permitted, fsync",
         "stack":"Error: EPERM: operation not permitted, fsync
           at fsyncSync (node:fs:1285:11)
           at fsyncPath (.../out/daemon/index.mjs:158447:5)
           at writeConfigFields (.../out/daemon/index.mjs:158466:5)
           at persistCustomMcpSnapshot (.../out/daemon/index.mjs:158424:5)
           at saveConfigSnapshot (.../out/daemon/index.mjs:158396:3)
           at loadConfig (.../out/daemon/index.mjs:158262:5)"},"msg":"Daemon failed"}
[warn]  [linkcode/desktop] daemon exited (code 1); restarting
...
[error] [linkcode/desktop] daemon keeps exiting (last code 1); giving up
```

## Root cause

`fsyncPath()` opens the file **read-only** and then fsyncs it:

```js
function fsyncPath(path) {
  const descriptor = openSync(path, "r");   // <-- read-only handle
  try {
    fsyncSync(descriptor);
  } finally {
    closeSync(descriptor);
  }
}
```

On Windows, `fsync` on a read-only handle fails (`ERROR_ACCESS_DENIED` mapped to `EPERM`). I verified this locally: fsync on a handle opened with `"r"` fails deterministically, while fsync on a handle opened with `"r+"`/`"w"` succeeds on the same files and directory. The directory fsync in `writeConfigFields` is already guarded with `if (process.platform !== "win32")`, but the temp-file fsync (`fsyncPath(temporaryPath)`) is not.

Trigger chain: accounts stored with inline credentials → `loadConfig` runs the migration → `saveConfigSnapshot` writes the vault, then `writeConfigFields` writes config.json → `fsyncPath(temp)` throws EPERM → daemon exits before `renameSync` → config.json still contains the inline credential → next startup runs the migration again → infinite crash loop.

## Workaround (for affected users)

Complete the migration by hand:

1. Quit LinkCode.
2. Back up `%USERPROFILE%\.linkcode\config.json` and `secrets.json`.
3. In `config.json`, remove the `key`/`token` from each account's `credential` but keep the `credential` object shape, e.g. `"credential": {"type": "api-key"}` (the daemon re-attaches the secret from the vault by account id).
4. Restart LinkCode.

## Suggested fix

- Open the fsync descriptor with write access (`openSync(path, "r+")`), or fsync the temp file's own write descriptor before `renameSync`, or skip `fsyncPath` on win32 (the rename-based write is already durable enough for this use case).
- Optionally: make the migration tolerant of a failed fsync (treat EPERM on win32 as non-fatal) so the vault/config state stays consistent and the daemon still starts.

### LINKED ISSUE #438: [Bug] pi agent catalog fails: 'model must be provider/modelId' when account default model is a bare id; SDK missing (ERR_MODULE_NOT_FOUND) until managed runtime ready
## Summary

The `pi` agent catalog fails to load when the active account's default model is a bare model id (e.g. `deepseek-v4-flash`), because the pi adapter requires `provider/modelId` format and throws, which fails the whole catalog. Separately, at startup the daemon reports `@earendil-works/pi-coding-agent` as missing (ERR_MODULE_NOT_FOUND) until the managed runtime asset is ready, producing repeated "Failed to load agent catalog" noise.

## Environment

- LinkCode desktop 0.20.0
- Windows 11 x64
- Account created from the "opencode-go" service catalog: `{ "model": "deepseek-v4-flash", "endpoint": { "baseUrl": "https://opencode.ai/zen/go/v1", "protocol": "openai-chat" } }`

## Symptom

`%APPDATA%\LinkCode\logs\main.log`:

```
{"subsystem":"agent","operation":"agent.catalog","err":{"type":"Error",
 "message":"pi: model must be 'provider/modelId' (got 'deepseek-v4-flash')",
 "stack":"Error: pi: model must be 'provider/modelId' (got 'deepseek-v4-flash')
   at createConfiguredRegistry (.../out/daemon/index.mjs:97941:11)
   at PiAdapter.startCatalog (.../out/daemon/index.mjs:97993:17)"},
 "msg":"Failed to load agent catalog"}
```

and, during the first ~20 s after startup:

```
"message":"pi: SDK '@earendil-works/pi-coding-agent' is unavailable (Error: Cannot find package '@earendil-works/pi-coding-agent' imported from .../out/daemon/index.mjs)"
"cause":{"code":"ERR_MODULE_NOT_FOUND"}
```

## Root cause 1 — model format

`PiAdapter.startCatalog` → `createConfiguredRegistry` does:

```js
const ref = opts.model ? parseModel(opts.model) : null;
if (!ref && opts.model) {
  throw new Error(`pi: model must be 'provider/modelId' (got '${opts.model}')`);
}
```

`opts.model` is the account's default model, which service-catalog accounts store as a bare id (the correct format for the claude-code/codex/opencode CLIs that share the same account). A bare id makes pi's catalog throw, so the pi agent surface is entirely unavailable until the user manually sets the model to `provider/modelId` (e.g. `opencode-go/deepseek-v4-flash`).

## Root cause 2 — SDK not bundled

`PiAdapter.importSdk()` first tries `agentRuntimeProber.resolveEntry("pi")` (the managed runtime asset) and only falls back to `import("@earendil-works/pi-coding-agent")`, which is not bundled in `app.asar`. Catalog retries fail with ERR_MODULE_NOT_FOUND until `asset.ensure` finishes ("Managed agent runtime ready", ~20 s after startup). If the managed runtime is GC'd by `asset.gc` (we observed "Removed superseded managed assets"), the errors return until the asset is re-ensured.

## Repro

1. Create an account from a service catalog with a bare default model (e.g. `deepseek-v4-flash`).
2. Enable the `pi` provider.
3. Restart the app; watch `agent.catalog` errors in `main.log`.

## Workaround

Set the account's model to `provider/modelId` format (e.g. `opencode-go/deepseek-v4-flash`) in the app settings or `~/.linkcode/config.json`.

## Suggested fix

- pi adapter: tolerate bare model ids by defaulting the provider to the account's service/endpoint name instead of throwing (or normalize the account model when configuring pi), so a shared account with a bare model id still produces a working catalog.
- Packaging: bundle `@earendil-works/pi-coding-agent` in the app, or gate catalog errors on the managed-runtime readiness state to avoid the misleading ERR_MODULE_NOT_FOUND noise.

lucas77778:
## Summary

* fsync daemon config writes through the temporary file's writable descriptor, preventing the Windows credential-migration crash loop
* resolve bare Pi model ids from the resumed provider, known provider, or a unique model/endpoint match
* pause Pi catalog requests until its managed runtime is available, avoiding startup module-not-found noise

## Verification

* `pnpm check:ci`
* `pnpm test` (2956 passed, 1 skipped)
* regression coverage for `opencode-go/deepseek-v4-flash`

Closes arcboxlabs/linkcode#437<br>Closes arcboxlabs/linkcode#438

### DIFF
--- a/apps/daemon/src/__tests__/config-persistence.test.ts
+++ b/apps/daemon/src/__tests__/config-persistence.test.ts
@@ -15,10 +15,10 @@ import { createProviderConfigStore } from '../provider-store';
 import { createInMemoryVault } from './fixtures/in-memory-vault';
 
 const fsMocks = vi.hoisted(() => ({
-  openTargets: new Map<number, string>(),
+  openTargets: new Map<number, { path: string; flags: string }>(),
   renameTarget: null as string | null,
   renameTargets: [] as string[],
-  syncTargets: [] as string[],
+  syncTargets: [] as Array<{ path: string; flags: string }>,
 }));
 
 vi.mock('node:fs', async (importOriginal) => {
@@ -33,7 +33,7 @@ vi.mock('node:fs', async (importOriginal) => {
       }
     },
     fsyncSync(descriptor: number) {
-      fsMocks.syncTargets.push(fsMocks.openTargets.get(descriptor) ?? '');
+      fsMocks.syncTargets.push(fsMocks.openTargets.get(descriptor) ?? { path: '', flags: '' });
       actual.fsyncSync(descriptor);
     },
     openSync(
@@ -42,7 +42,7 @@ vi.mock('node:fs', async (importOriginal) => {
       mode?: Parameters<typeof actual.openSync>[2],
     ) {
       const descriptor = actual.openSync(path, flags, mode);
-      fsMocks.openTargets.set(descriptor, String(path));
+      fsMocks.openTargets.set(descriptor, { path: String(path), flags: String(flags) });
       return descriptor;
     },
     renameSync(
@@ -115,9 +115,12 @@ describe('provider config persistence', () => {
     expect(store.getAccounts()).toEqual([oauthAccount]);
     expect(statSync(config).mode & 0o777).toBe(0o600);
     expect(fsMocks.renameTargets).toEqual([config]);
-    expect(fsMocks.syncTargets[0]).toContain(join(dir, '.config.'));
-    expect(fsMocks.syncTargets[0]?.endsWith('.tmp')).toBe(true);
-    expect(fsMocks.syncTargets.slice(1)).toEqual(process.platform === 'win32' ? [] : [dir]);
+    expect(fsMocks.syncTargets[0]?.path).toContain(join(dir, '.config.'));
+    expect(fsMocks.syncTargets[0]).toMatchObject({ flags: 'wx' });
+    expect(fsMocks.syncTargets[0]?.path.endsWith('.tmp')).toBe(true);
+    expect(fsMocks.syncTargets.slice(1).map(({ path }) => path)).toEqual(
+      process.platform === 'win32' ? [] : [dir],
+    );
   });
 
   it('rejects corrupt JSON without replacing the file or publishing memory', () => {
--- a/apps/daemon/src/config.ts
+++ b/apps/daemon/src/config.ts
@@ -457,13 +457,14 @@ function writeConfigFields(
   mkdirSync(directory, { recursive: true });
   const temporaryPath = join(directory, `.config.${process.pid}.${randomUUID()}.tmp`);
   try {
-    writeFileSync(temporaryPath, `${JSON.stringify(file, null, 2)}\n`, {
-      encoding: 'utf8',
-      flag: 'wx',
-      mode: 0o600,
-    });
-    chmodSync(temporaryPath, 0o600);
-    fsyncPath(temporaryPath);
+    const descriptor = openSync(temporaryPath, 'wx', 0o600);
+    try {
+      writeFileSync(descriptor, `${JSON.stringify(file, null, 2)}\n`, { encoding: 'utf8' });
+      chmodSync(temporaryPath, 0o600);
+      fsyncSync(descriptor);
+    } finally {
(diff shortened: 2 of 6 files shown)


PeronGH (approved):
Reviewed the three production changes. The fsync-through-the-writable-handle fix is correct: the old `fsyncPath` reopened the temp file read-only, so the sync targeted a different descriptor than the one that wrote the bytes. Write/chmod/fsync/close ordering and the post-rename directory fsync are preserved. Pi bare-model resolution correctly refuses ambiguous endpoint matches, and gating the Pi catalog on runtime availability leaves the other four agents untouched. CI green.

PR #456, opened 2026-08-17
feat(providers): add StepFun API support

lucas77778:
## Summary
- add a dedicated StepFun API provider for the China endpoint at `api.stepfun.com`
- route Claude Code through Anthropic Messages, Codex through the officially supported OpenAI Responses endpoint, and OpenCode through its models.dev-backed OpenAI Chat provider; Pi follows the existing unknown-provider fallback over StepFun's Anthropic-compatible endpoint
- add StepFun model discovery, China-console setup copy, the official lobe-icons StepFun color glyph, and provider/binding tests

## Testing
- `pnpm check:ci`
- `pnpm test` (2958 passed, 1 skipped)
- exercised the provider flow in the Electron desktop and verified the StepFun tile, `platform.stepfun.com` hint, `sk-…` key field, protocol list, and official gradient glyph

## Vendor reference
- [StepFun Responses API](https://platform.stepfun.com/docs/zh/api-reference/responses/responses-create): `POST https://api.stepfun.com/v1/responses`, currently supporting `step-3.7-flash`

Closes [CODE-571](https://linear.app/arcbox/issue/CODE-571/featproviders-add-stepfun-support)


packages/presentation/i18n/src/locales/en.ts
```diff
@@ -1130,6 +1131,7 @@ export const en = {
         'openai-api': 'A key from platform.openai.com',
         xai: 'A key from console.x.ai',
         deepseek: 'A key from platform.deepseek.com',
+        stepfun: 'A key from platform.stepfun.ai',
```
PeronGH: Both locales' key hints point at the international console (`platform.stepfun.ai`; `zh-cn.ts:1107` too), but every endpoint in this catalog entry is wired to `api.stepfun.com` — StepFun's separate China platform. The two are distinct account systems: a key from `platform.stepfun.ai` does not authenticate against `api.stepfun.com`. Either add `api.stepfun.ai` variants for international keys or point both hints at `platform.stepfun.com`. As is, an English-speaking user follows the hint and gets auth failures.

packages/foundation/providers/src/catalog.ts
```diff
@@ -133,6 +135,24 @@ export const SERVICE_CATALOG: ServiceDescriptor[] = [
...
   },
+  {
+    id: 'stepfun',
+    label: 'StepFun',
+    group: 'direct',
+    kind: 'endpoint',
+    credentialType: 'api-key',
+    variants: {
+      anthropic: { baseUrl: 'https://api.stepfun.com' },
+      'openai-responses': { baseUrl: 'https://api.stepfun.com/v1' },
```
PeronGH, line 145: StepFun has no Responses API — its OpenAI-compatible surface is chat completions only (the vendor's own Codex guide uses `wire_api = "chat"`). Codex ≥ 0.122 speaks Responses exclusively (`resolve.ts` documents chat as a config-load error), so this variant can never back a codex session: resolution reports `native` and every request then hits a dead `/v1/responses`. Drop this variant and remove `codex` from `supportedAgents` so codex resolves to `protocol-unsupported`.

PeronGH (changes requested):
Requesting changes — two issues found while verifying the StepFun wire facts against vendor docs:

1. **Wrong wire for codex.** StepFun serves Anthropic-messages (root) and OpenAI chat (`/v1`) only — no Responses API. Since codex ≥ 0.122 speaks Responses exclusively, the `openai-responses` variant resolves codex to a dead endpoint. Drop the variant and remove `codex` from `supportedAgents`.
2. **Region mismatch.** Every endpoint is the China platform (`api.stepfun.com`) while both locales' key hints point at `platform.stepfun.ai` — separate account systems, so an international key can't authenticate here. Wire `api.stepfun.ai` variants or fix the hints.

The claude-code (root Anthropic URL) and opencode (`knownProvider: 'stepfun'`, per models.dev) parts check out.

packages/foundation/providers/src/catalog.ts
lucas77778: Rechecked against the current official API reference before changing this. StepFun now explicitly documents OpenAI Responses compatibility at `POST https://api.stepfun.com/v1/responses`, currently for `step-3.7-flash`: https://platform.stepfun.com/docs/zh/api-reference/responses/responses-create. Codex uses exactly that wire, so removing the variant would discard a supported integration; the Responses variant and Codex coverage are retained, and the vendor reference is now linked in the PR body.

packages/presentation/i18n/src/locales/en.ts
lucas77778: Fixed in f9530ee5. Both locale hints now point to `platform.stepfun.com`, matching the China API host.

lucas77778:
Review feedback addressed in f9530ee5:

- corrected both key hints to the China console, `platform.stepfun.com`
- replaced the hand-written glyph with `~icons/lobe-icons/stepfun-color` and restored the original glyph-map type
- removed the StepFun-only `supportedAgents` axis and resolver gate; Pi now follows the package's existing unknown-provider fallback, so the inaccurate reason and documentation mismatch disappear
- added the missing `sk-…` secret placeholder and simplified the StepFun resolver test
- retained Codex after revalidating the current official StepFun reference, which explicitly documents OpenAI Responses compatibility at `POST /v1/responses` for `step-3.7-flash`: https://platform.stepfun.com/docs/zh/api-reference/responses/responses-create

Verification: `pnpm check:ci`; `pnpm test` (2958 passed, 1 skipped); Electron UI verified for the official gradient glyph, `.com` hint, key placeholder, and protocol list.

@PeronGH please re-review when convenient.

packages/foundation/providers/src/catalog.ts
```diff
@@ -133,6 +133,23 @@ export const SERVICE_CATALOG: ServiceDescriptor[] = [
...
```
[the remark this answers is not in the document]
lucas77778: Added the matching Responses model constraint in 0b8105a2.

packages/foundation/providers/src/catalog.ts
PeronGH: Withdrawn — rechecked against StepFun's current API reference: `POST /v1/responses` is officially documented (OpenAI-Responses-compatible, `step-3.7-flash`). The variant and codex coverage are correct; the new constraint comment covers the model limitation. My original comment was based on the older Step-3.5-Flash README, which predates the Responses API.

PeronGH (approved):
LGTM. Verified against StepFun's current API reference: `/v1/responses` is served for `step-3.7-flash`, matching the `openai-responses` variant and its constraint comment; both locale hints now match the China endpoints. The claude-code root URL and opencode `knownProvider` (`stepfun` in models.dev) also check out. Could not run the test suite in this environment — CI is the gate.

PR #458, opened 2026-08-18
fix(workbench): require models when adding providers

lucas77778:
## Summary

- require at least one selected model before manually adding OAuth, catalog API, gateway, or custom endpoint providers
- persist the selected models on every newly created account and automatically probe/select all models returned by LinkCode Gateway
- keep local auto-detected providers unchanged: they already select every model in the harness's curated model set

Linear: [CODE-589](https://linear.app/arcbox/issue/CODE-589/fixworkbench-require-models-when-adding-providers)

## Verification

- `pnpm check:ci`
- `pnpm test` — 2964 passed, 1 skipped
- provider add-flow tests — 15 passed
- exercised the provider form in the running webview and recorded the empty-selection validation followed by successful model selection and provider creation

## Checklist

- [x] If a wire message changed: `WIRE_PROTOCOL_VERSION` is bumped (not applicable; no wire change)
- [x] Docs and comments are updated where behavior changed (not applicable; UI copy and tests updated)

### DIFF
--- a/packages/client/workbench/src/settings/providers/__tests__/add-flow.test.tsx
+++ b/packages/client/workbench/src/settings/providers/__tests__/add-flow.test.tsx
@@ -5,6 +5,7 @@ import { act, cleanup, fireEvent, render, screen, waitFor } from '@testing-libra
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import type { AgentRuntimeOnboarding } from '../../../agent-runtime/onboarding';
 import { AddAccountForm, ServiceCatalogView } from '../add-flow';
+import type { ModelSources } from '../model-selection';
 
 function translateKey(key: string): string {
   return key;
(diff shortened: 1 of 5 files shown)


PeronGH (approved):
Verified the constraint is UI-only: `models` stays `z.array(AccountModelSchema).optional()` in `packages/foundation/schema/src/model/account.ts`, so existing zero-model accounts remain valid and no wire/schema change is implied. Edit mode correctly keeps the looser `CustomDraftSchema`, so the requirement can't retroactively block editing an old account.

Traced the gateway path against `linkcodehq`, since it's the one flow that gains a network dependency mid-creation: `modelListSource('linkcode-gateway')` resolves to `gateway.linkcode.ai/v1/models`, which serves the compile-time `MODEL_CATALOG` constant (75 models across four bundled provider manifests) with no DB read or upstream call, and auth is a strongly-consistent better-auth `verifyApiKey` against the same DB `createKey` just wrote to. So the probe can't return empty and has no failure mode `createKey` doesn't already have — the `models.length === 0` throw is a safety net rather than a reachable path. Caching the minted key in `createdKey` so a retry doesn't mint a second one is the right call.

One non-blocking note: a network failure strictly between `createKey` and the probe response leaves a gateway key minted with no account attached if the user then closes the dialog. Pre-existing shape of the flow, slightly widened — worth a follow-up only if orphaned keys show up in practice.

PR #459, opened 2026-08-19
feat(providers): allow ordering accounts and selecting default models

lucas77778:
## Summary

- add drag-and-drop priority ordering for provider accounts and persist the complete account order
- update New Task defaults from the first compatible account's first model
- allow users to promote a selected account model to the default position
- add English/Chinese copy and focused tests

Linear: [CODE-590](https://linear.app/arcbox/issue/CODE-590/featproviders-allow-ordering-accounts-and-selecting-default-models)

## Verification

- `pnpm check:ci`
- `pnpm test` — 2966 passed, 1 skipped
- exercised the mock Webview end-to-end and verified that dragging DeepSeek above Anthropic changes the New Task default from Claude Opus 5 to DeepSeek V4 Pro

## Checklist

- [x] No wire message changed
- [x] New code and assets are my own work

### DIFF
--- a/packages/client/workbench/src/settings/providers/__tests__/model-selection.test.tsx
+++ b/packages/client/workbench/src/settings/providers/__tests__/model-selection.test.tsx
@@ -7,8 +7,9 @@ import { useState } from 'react';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { ModelSelection } from '../model-selection';
 
-function translateKey(key: string): string {
-  return key;
+function translateKey(key: string, values?: Record<string, unknown>): string {
+  const interpolation = values ? Object.values(values).join(',') : '';
+  return interpolation ? `${key}:${interpolation}` : key;
 }
 
 vi.mock('use-intl', () => ({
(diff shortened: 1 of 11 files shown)


PeronGH (dismissed):
Checked the ordering semantics against the contract rather than the diff alone: `packages/foundation/providers/src/enabled-models.ts` already defines the default as "each enabled account in pool order, contributing its picked set in its own order — the first entry is the agent's default", and the daemon reads the same thing (`provider-config.ts`, `enabledAccounts(...)[0]`). Both halves of this PR edit exactly those two orderings instead of introducing a stored default, which is the right shape.

Persistence path looks correct — optimistic `mutateAccounts(..., { revalidate: false })` → `trigger` → bare `mutateAccounts()` reconcile, with revert + toast on failure. The `pool` captured in the `catch` is the pre-drag order, so the revert restores the right thing. dnd-kit usage avoids the known traps: `useSortable` results are destructured into plain identifiers before reaching `ref=`, and a flat single list correctly needs no `onDragOver` + `preventDefault`.

Non-blocking notes:

- `providers-settings.tsx` — `if (reordered.length !== pool.length) return;` is unreachable. `providerAccountListViewModel` is a 1:1 `.map` with no filtering and `accountsById` is built from the same `pool`, so the lengths can't diverge. A silent return there would also read to the user as the same unexplained snap-back that the error path now avoids.
- `account-master-list.tsx` — the no-op guard (`reordered.some((id, index) => id !== current[index])`) is what stops a drag that lands where it started from writing to the daemon, and it's untested. With `move` already mocked, it's one `mockReturnValueOnce`. The `event.canceled` early return is uncovered too.
- `model-selection.test.tsx` — "keeps only the disabled default marker fully opaque" asserts `.className` contains the literal `disabled:opacity-100`. It covers the `isDefault &&` branch, but by restating the implementation's own string, and jsdom computes no Tailwind so the opacity itself isn't verified. The sibling test (default moves to head) is the one carrying the weight.
- The default-model star is a `disabled` button used as a state indicator. Native `disabled` drops it from the tab order, so a keyboard user infers the default from the absence of a "Make X default" button rather than being told. `aria-pressed` on an enabled no-op would state it directly.

PeronGH (approved):

PR #460, opened 2026-08-20
fix(workbench): hide disabled harnesses from new tasks

lucas77778:
## Summary

Fixes [CODE-597](https://linear.app/arcbox/issue/CODE-597/fixworkbench-hide-disabled-harnesses-from-new-tasks).

- Filter the new-task Harness picker using the enablement state from Agents settings, so a disabled OpenCode Harness is no longer offered.
- Wait for provider configuration before exposing choices, fall back when the remembered Harness is disabled, and prevent submission when no Harness is enabled.
- Pass the filtered Harness list through both the shared and desktop shell paths, with regression coverage for filtering and fallback behavior.

## Verification

- `pnpm check:ci`
- `pnpm test` — 2973 passed, 1 skipped
- Ran the Electron desktop surface, disabled OpenCode under Agents, and confirmed OpenCode disappeared from the new-task Harness picker; restored the setting afterward.

## Checklist

- [x] If a wire message changed: `WIRE_PROTOCOL_VERSION` is bumped

### DIFF
--- a/apps/desktop/src/renderer/src/shell/desktop-shell.tsx
+++ b/apps/desktop/src/renderer/src/shell/desktop-shell.tsx
@@ -79,6 +79,7 @@ export function DesktopShell({
   runtimeCues,
   attachmentSupport,
   agentCatalogs,
+  selectableHarnesses,
   accountModels,
   newSessionPreferredEfforts,
   newSessionPreferredBranches,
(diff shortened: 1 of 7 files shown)


PeronGH (approved):
Reviewed: filtering, fallback, and blocked-submission behavior are correct and well-scoped. Wired through all three shell render sites; empty/undefined harness handled safely in Composer and ModelSelectorMenu. Approving.

PR #479, opened 2026-08-25
feat(client): consume brand identity agents/services allowlists

lucas77778:
## Summary

Client half of per-brand agent/provider control (CODE-618, part of CODE-612). The pinned render pipeline may now deliver optional `agents` / `services` allowlists on the config build bundle (publisher side: arcboxlabs/linkcodehq#49); this PR consumes them at build time. Absent fields = unrestricted — the standard build is byte-identical.

- **Contract**: `ConfigBuildBundleSchema` and the brand-identity mirror accept optional `agents` (AgentKind subset) / `services` (kebab-case ids); non-empty, duplicate-free, still fail-closed on unknown fields. Frozen fixtures untouched. Both files avoid a *value* import of `@linkcode/schema`'s root barrel (only reachable from build scripts under plain Node, which can't resolve its directory re-export) — the known agent-kind list is duplicated locally instead, matching this codebase's `CONFIG_PLATFORMS`/`CONFIG_CHANNELS` precedent.
- **Desktop main**: new `MAIN_VITE_AGENT_RESTRICTIONS` define → `agent-restrictions.ts` (present-but-invalid aborts boot, mirroring brand identity), surfaced to the renderer via a synchronous `identity.restrictions()` SystemBridge channel that **fails closed** if the main-process snapshot is ever missing (never defaults to unrestricted).
- **Daemon enforcement boundary (agents)**: `LINKCODE_ALLOWED_AGENTS` from the supervisor is checked wherever an adapter process could actually start: `SessionOrchestrator.startLive` (the single chokepoint every start/resume/relaunch path constructs a live adapter through) and `agent.catalog` in `AgentRequestHandler` (whose `startCatalog` spawns real agent processes for codex/opencode). This refuses only a *new* live run of an excluded kind; it never touches `session.list`'s per-record `historyCapabilities` or `HistoryService`'s list/read/branch-cursor paths, so a persisted session of an excluded kind stays fully listable and readable. The boot runtime-probe snapshot and `collectAgentRuntimes()` are filtered (`filterAgentRuntimes`) so an excluded agent always reads `missing` regardless of how it was actually detected, and the wire-level asset surface is gated (`restrictedAssetService`): `asset.ensure` refuses an excluded agent's managed download and `statuses`/`subscribe` omit its entries entirely — both live in `apps/daemon/src/agent-restrictions.ts`, composed at the daemon boundary without touching `@linkcode/agent-adapter`'s public API.
- **Daemon enforcement boundary (services)**: catalog-display filtering only (UI layer) — as planned, there is no daemon-side enforcement for `services`.
- **Packaging**: `AGENT_SDK_PACKAGE_PATHS` is the single source of truth for both brand-conditional electron-builder `files` excludes (temp `extends` overlay, written into a fresh `mkdtempSync` dir per run) and a new `verifyNoRestrictedAgentPackages` post-pack asar assertion (exact path-segment matching).
- **UI**: harness picker intersects the brand allowlist and its *submenu* disappears at ≤1 remaining agent (desktop composer + mobile new-thread sheet) — the trigger's agent icon stays visible whenever any harness is selectable at all, since it's the only harness signal once the submenu is gone. Add-account catalog filters oauth entries by bound agent and endpoint/`custom` entries by service id. `serviceById`, account list/detail, and resolution stay unfiltered so stored accounts keep rendering even if their service is no longer offered.

## Fixed after review / CI

- CI's `Desktop App Entry` / `e2e:unpackaged` failed with `ERR_UNSUPPORTED_DIR_IMPORT` loading `vite.main.config.mts` under plain Node — fixed by the contract-layer import change above.
- `SessionOrchestrator`'s adapter-factory wrapper originally threw for any excluded kind at construction, which broke `session.list` for daemons with even one persisted session of an excluded kind — moved to the `startLive`-only boundary described above.
- Composer's harness icon incorrectly disappeared at ≤1 harness (should only hide the submenu) — reverted to `harnesses.length > 0` for the icon.
- Renderer `identity.restrictions()` defaulted to unrestricted on a missing IPC snapshot — now throws.
- Minor cleanups: unreachable branch in `daemonAllowedAgents()`, an ambient-env guard now consistent with its neighbors, and the packaging files-exclude overlay now uses a unique temp dir instead of a fixed filename.

## Validation

- `pnpm typecheck` / `pnpm lint` / `pnpm format:check` clean
- `pnpm test`: 3051–3052 passed; 3 failures confirmed pre-existing on the unmodified base (tmpdir/realpath quirk in `release-artifact.test.ts`, unrelated to this PR); one flaky timing failure observed intermittently in `dev-mock-transport.test.ts` (an unrelated pre-existing integration test, polling-based, flaky under the sandbox's concurrent load — passes cleanly in isolation)
- `cd apps/desktop && pnpm e2e:unpackaged`: daemon/main/preload/renderer build and the full Electron IPC boundary read (`linkcodeSystem`/`linkcodeConfig`) now succeed; the only remaining failure locally is the script's hardcoded `assert.equal(boundary.platform, 'linux')`, pre-existing and Linux-CI-only (this sandbox is macOS)
- TODO (post-merge, with the restricted brand registered in linkcode-config): package a restricted installer and run `verify-artifacts.mts` both ways; e2e config-canary restricted fixture

Linear: CODE-618


xiaoland (commented):
LGTM. Though agent-login.start still bypasses the restricted-agent spawn boundary, but I see a same comment being marked as resolved.

xiaoland (dismissed):

xiaoland (approved):
I see you resolve the conflicts by merging the main branch, and nothing else changed. so, LGTM

PR #480, opened 2026-08-25
Qi/code 622

adminliu-main:
## Summary

新增 LinkCode 插件市场：桌面端现在可以浏览官方插件目录、安装插件，并通过
manifest 驱动的设置表单配置插件。关联 CODE-622。

- **Daemon** —— 新增 `marketplace/` 服务（目录拉取与刷新）和
  `plugin-store/`（暂存式安装、失败自动回滚；插件设置项中的 secret 值写入
  vault），两者均接入控制通道。
- **客户端** —— 控制通道新增市场相关操作（浏览 / 安装 / 配置），SDK 增加
  对应 operations，workbench 提供 SWR hooks。
- **设置界面** —— 插件设置新增 LinkCode 页签：目录卡片展示安装状态（含
  "本地已装更新版本"识别）；单插件配置弹窗（secret 字段不回显折叠默认值；
  与默认值相等的设置项按删除持久化）。
- **Wire 协议** —— 纯新增帧，`WIRE_PROTOCOL_VERSION` 78 → 80，
  `MIN_COMPATIBLE_WIRE_VERSION` 不变。
- **mail 插件移出本仓库** —— 改由官方插件市场仓库分发
  （arcboxlabs/linkcode-plugins-official#1），本 PR 不再内置任何插件；
  本地调试用 `scripts/dev-marketplace.mts` fixture。

## Verification

- `pnpm check:ci` —— 通过（format / lint / typecheck，0 错误）
- `pnpm test` —— 通过（仅有的 3 个失败是 release-artifact 的已知环境问题，
  master 上同样失败，与本改动无关）
- `apps/daemon/e2e/plugin-marketplace.e2e.ts` —— 通过：基于 dev marketplace
  fixture 端到端跑通目录拉取、暂存安装、设置写入（secret 进 vault）与回滚
- 渲染层由组件/视图测试覆盖（`linkcode-config-dialog`、`linkcode-config`、
  `view`、`plugin-market`）；设置界面可用 `scripts/dev-marketplace.mts`
  在本地实际驱动
- 三轮 review 意见已在 `6373d0cd`、`05b49deb` 中处理完毕

## Checklist

- [x] `pnpm check:ci` 和 `pnpm test` 均通过（无 Rust 改动）
- [x] 已运行受影响的界面/服务并确认改动生效（daemon e2e + dev marketplace
  fixture）
- [x] wire 消息有变更：`WIRE_PROTOCOL_VERSION` 已提升（78 → 80，纯新增，
  兼容地板不变）
- [x] 新增代码与资产为本人工作成果，或已在上方注明来源与许可兼容性
- [x] 行为变更处的文档与注释已更新（新增环境变量已写入
  `docs/ENVIRONMENT.md`）


apps/daemon/src/plugin-store/store.ts
```diff
@@ -0,0 +1,586 @@
...
+        pluginId,
+        path: targetDir,
+        retiredDir,
+        operation: 'plugin.install.rollback-publish',
+      },
+      'Failed to remove a published plugin after registry persistence failed',
+    );
+    return;
+  }
+  if (retiredDir !== undefined) restorePluginPackage(retiredDir, targetDir, pluginId);
```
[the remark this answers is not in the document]
adminliu-main: Addressing the Greptile summary — I traced both concerns through the code:

**"stale registry metadata paired to new package contents"** — the only metadata that can go stale in the publish→persist window is `InstalledLinkCodePlugin.integrity` (`packages/foundation/schema/src/model/linkcode-plugin.ts:318`). That field has zero consumers anywhere in daemon/engine/client: update detection compares `version` (`comparePluginVersions` in `packages/client/workbench/src/settings/plugins/view.ts`), not integrity. It's recorded purely as "digest verified before publish" for future use, so a stale value can't change any listing, startup, or update decision.

**"make the registered package unavailable"** — a failed restore leaves the package at its `.tmp-retired-*` path with an empty target. Boot reconciliation recovers exactly that shape: `reconcileRetiredPackage` (`apps/daemon/src/plugin-store/store.ts`) sees `hasExactRecord` + missing target and renames it back (`restored`). So it's transient, not a permanent unavailable state.

**On the crash window itself** — real, but bounded and now explicitly documented: `installExclusive` carries a comment marking the publish→persist gap as a known, harmless window (orphaned version dir / stale integrity, both reclaimed by the next install of that version).

So this reads as a low-severity crash-consistency nuance rather than a merge blocker. If the reasoning changes anything, a re-score would be appreciated.

packages/host/engine/src/session/start-options-resolver.ts
```diff
@@ -165,6 +178,66 @@ export class SessionStartOptionsResolver {
...
+          }
+        }
+      }
+      return { options, warnings };
+    }
+    const servers = [...(options.mcpServers ?? [])];
+    for (const entry of entries) {
+      const settings = store.getSettings(entry.installed.id);
+      for (const component of entry.manifest.components) {
+        if (component.kind !== 'mcp-server') continue;
```
lucas77778, line 235: **[blocking] The marketplace currently allows skill-only releases to be installed, but this path skips every component other than `mcp-server`.** For example, this PR's `linkcode/notes` mock will report a successful install and appear in the installed list, yet there is no path in the repository that projects its `SKILL.md` into an agent; `manifest.assets` is likewise never consumed. If skill/asset projection is outside this PR's scope, filter out or disable these releases at the catalog/install boundary rather than reporting a successful install that provides no functionality.

apps/daemon/src/plugin-store/store.ts
```diff
@@ -0,0 +1,586 @@
...
+  ): Promise<InstalledLinkCodePluginEntry> {
+    return this.serialize(release.manifest.id, () => installExclusive(release, marketplaceId));
+  }
+
+  uninstall(pluginId: string): Promise<void> {
+    return this.serialize(pluginId, () => {
+      const records = readRegistry();
+      const matches = records.filter((entry) => entry.id === pluginId);
+      if (matches.length > 0) {
+        for (const record of matches) rmSync(record.path, { recursive: true, force: true });
```
lucas77778: **[blocking] Uninstall recursively deletes the live package before persisting the registry, and cannot roll back on failure.** If `writeRegistry` then fails because of ENOSPC/EACCES, the registry still points to a directory that has already been deleted; a failure partway through recursive deletion can also leave a partial package. In the other direction, if config or vault cleanup fails after the registry commit, the plugin is already invisible but its config/secrets remain indefinitely and will be inherited by a later reinstall. Please retire the package first and restore it on registry failure, as install does, and use a recoverable tombstone/journal to retry config and vault cleanup.

apps/daemon/src/plugin-store/store.ts
```diff
@@ -0,0 +1,586 @@
...
+  }
+}
+
+function readRegistry(): InstalledLinkCodePlugin[] {
+  const path = pluginRegistryPath();
+  let raw: string;
+  try {
+    raw = readFileSync(path, 'utf8');
+  } catch {
+    return [];
```
lucas77778, line 738: **[blocking] Only `ENOENT` should mean a first-run empty registry.** This treats every read error as `[]`; malformed JSON, a non-array root, and invalid records below are degraded in the same way. The next install's `upsertRegistry()` will then write a file containing only the new plugin, silently dropping every existing installation record. Non-ENOENT read errors and parse/validation failures should block mutations rather than letting a degraded read participate in an overwrite. The persisted `record.path` should also be validated against `pluginPackageDir(id, version)`, or simply re-derived, so corrupted state cannot drive a recursive removal outside the Store.

packages/foundation/schema/src/model/linkcode-plugin.ts
```diff
@@ -106,6 +109,101 @@ const linkCodePluginSkillFields = {
...
+ * host can render a form without executing plugin code.
+ */
+export const LinkCodePluginSettingFieldSchema = z
+  .strictObject({
+    type: LinkCodePluginSettingTypeSchema,
+    label: z.string().min(1).optional(),
+    description: z.string().optional(),
+    secret: z.boolean().optional(),
+    required: z.boolean().optional(),
+    default: z.union([z.string(), z.number(), z.boolean()]).optional(),
```
lucas77778, line 134: **[blocking] `default` is not constrained by the setting `type`.** The current schema accepts `{type:'boolean', default:'false'}`, `{type:'number', default:true}`, and an enum default that is not one of its options. The store then folds these values in unchanged, causing the UI representation and the value injected at runtime to diverge. Please model the fields as a discriminated union by type (or validate the default type and enum membership in this refinement), and add these counterexamples to the tests.

apps/daemon/src/plugin-store/store.ts
```diff
@@ -0,0 +1,586 @@
...
+        const field = settings[fieldId];
+        if (field.secret) secretPatch.set(`${pluginId}/${fieldId}`, undefined);
+        else {
+          const { [fieldId]: _removed, ...rest } = nextNonSecret;
+          nextNonSecret = rest;
+        }
+      }
+    }
+    if (patch.set) {
+      for (const [fieldId, value] of Object.entries(patch.set)) {
```
lucas77778: **[blocking] The daemon does not validate config writes against the manifest.** The wire only guarantees that each value is a primitive, so an older or alternate client can write a string to a boolean field or any value to an enum; this code persists it successfully (`secret` values are also unconditionally stringified). Validate the complete post-patch state against the current field schemas at the daemon boundary and reject invalid input as `invalid_request` rather than relying on the current UI. Upgrades also need to reconcile stored values against the new manifest: removing a field or changing its `secret` classification currently leaves the old value in the wrong config/vault location, and a non-secret-to-secret change leaves the value in plaintext in `config.json`.

packages/client/workbench/src/settings/plugins/linkcode-config.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/** Validate one raw form value against its declared field; `true` passes. */
+export function validatePluginConfigField(
+  field: LinkCodePluginSettingField,
+  raw: string | boolean,
+): true | PluginConfigFieldError {
+  if (field.type === 'boolean') return true;
+  const value = typeof raw === 'string' ? raw : String(raw);
+  if (value === '') {
+    // A blank secret is "keep the stored value", never an error.
```
lucas77778: **[blocking] For a newly installed plugin, a blank required secret does not mean 'keep the old value'.** The masked view returns neither the secret nor a presence bit indicating that the key is configured. A new plugin therefore has no old value, but this validation still passes, the patch omits the field, and the daemon does not validate the final required state; the session then starts without the corresponding environment variable. The wire/view should expose secret presence without exposing the value, the UI should interpret blank as keep only when a value is already configured, and the daemon should authoritatively validate the final required state.

packages/host/engine/src/session/start-options-resolver.ts
```diff
@@ -165,6 +178,66 @@ export class SessionStartOptionsResolver {
...
+      for (const entry of entries) {
+        for (const component of entry.manifest.components) {
+          if (component.kind === 'mcp-server') {
+            warnings.push({ serverName: component.name, reason: 'agent-unsupported' });
+          }
+        }
+      }
+      return { options, warnings };
+    }
+    const servers = [...(options.mcpServers ?? [])];
```
lucas77778, line 228: **[blocking] LinkCode MCP injection does not preflight names from native Codex provider plugins.** The custom-MCP path above calls `enabledMcpServerNames('codex')`, while this path checks only `options.mcpServers`. If an enabled native Codex plugin and a LinkCode plugin declare the same server name, the generated `mcp_servers.<name>.*` override replaces the native plugin's configuration without a `name-conflict` warning. Please reuse the same provider preflight here, including `provider-preflight-failed` behavior when discovery fails.

packages/client/workbench/src/settings/plugins/linkcode-config-dialog.tsx
```diff
@@ -0,0 +1,202 @@
...
+          <span className="font-medium text-sm">{label}</span>
+          {field.description === undefined ? null : (
+            <span className="text-muted-foreground text-xs">{field.description}</span>
+          )}
+        </div>
+        <Controller
+          control={control}
+          name={formKey}
+          render={({ field: switchField }) => (
+            <Switch
```
lucas77778, line 150: **[non-blocking / a11y] This Switch has no accessible name.** The visible `<span>` on the left is not a `<label>` and is not associated with the control by id, so a screen reader sees only an unnamed `switch` (the current test can likewise query it only with `getByRole('switch')`). Give the label and description stable ids, then set `aria-labelledby` / `aria-describedby` on the Switch.

lucas77778 (changes requested):
Overall implementation and tests are solid, and all existing CI checks pass. However, this version still has reachable paths that can report a successful but non-functional install, break registry/package consistency, or leave secrets behind after uninstall. The blocking issues are called out inline and should be addressed before merge; the Switch accessible-name note is explicitly non-blocking.

PR #485, opened 2026-08-27
fix(desktop): verify better-sqlite3's shipped prebuild

lucas77778:
The v0.27.0 release build failed on all three platforms with:

```
✗ mac/LinkCode.app/Contents/Resources: missing native binding node_modules/better-sqlite3/build/Release/better_sqlite3.node (better-sqlite3)
```

better-sqlite3 13 (#443) changed how the binding is distributed. v12 downloaded a per-Node-ABI
build into `build/Release/better_sqlite3.node` via `prebuild-install`; v13 ships eight N-API
(`NAPI_VERSION=10`) prebuilds in the tarball as `prebuilds/<platform>-<arch>.node`, and its
`binding.gyp` turns both targets into `type: none` whenever a prebuild exists. node-gyp therefore
only touches its stamps, `@electron/rebuild` is a no-op, and nothing is ever written to
`build/Release` — the packaged app is fine, `verify-artifacts.mts` was just looking in the old
place. Every other check in that run passed.

`verify-artifacts.mts` now asserts the target's prebuild instead, keeping the Mach-O/PE/ELF arch
check. Three stale comments about `prebuild-install` / `build/Release` are corrected with it.

## Verification

- The prebuild loads under Electron 43 (Node 24.18.0, N-API 10) and runs SQL — no rebuild needed.
- Packed an arm64 dev shell: `@electron/rebuild` stays a no-op and
  `app.asar.unpacked/node_modules/better-sqlite3/prebuilds/darwin-arm64.node` is in place.
- The updated check reports no problems for that app, and correctly reports the missing
  `darwin-x64.node` when pointed at an arm64 app as if it were the x64 one.
- Launched the packaged app: the daemon boots and migrates `daemon.db`, so the shipped prebuild is
  what it loads.
- `pnpm check:ci` passes.

## Follow-up (not in this PR)

v13's tarball carries all eight prebuilds and electron-builder smart-unpacks the whole module, so
every artifact now ships ~15 MB of foreign-platform `.node` files. Pruning the staging dir down to
the target (plus the host's, which is what keeps `binding.gyp` from compiling from source) removes
that; left out here to keep the release unblocking minimal.

### DIFF
--- a/apps/daemon/AGENTS.md
+++ b/apps/daemon/AGENTS.md
@@ -165,9 +165,9 @@ Runs via `tsx` in dev (`pnpm -F @linkcode/daemon dev`) and a `tsup` bundle in pr
   materializes a self-contained dir at `apps/daemon/standalone` (gitignored; pass an explicit path as
   argv for CI) via `pnpm --prod deploy` — the tsup bundle plus its runtime externals flat in the dir's
   own `node_modules`, runnable anywhere as `node --import ./dist/instrument.js dist/index.js`. This is
-  distinct from the desktop bundle: it targets **plain Node** (better-sqlite3 keeps its prebuild-install
-  binary — a **same-platform** artifact, build per target), and it prunes the host-arch agent CLI
-  platform packages (the daemon downloads them at runtime via `@linkcode/assets`, as the desktop
+  distinct from the desktop bundle: it targets **plain Node** (still a **same-platform** artifact — the
+  napi-rs optional deps and @sentry's profiler resolve host-only — build per target), and it prunes
+  the host-arch agent CLI platform packages (the daemon downloads them at runtime via `@linkcode/assets`, as the desktop
   does). The pi npm closure needs no prune (CODE-219): its SDK is a devDependency of
   agent-adapter, so the `--prod` deploy never materializes it — the daemon downloads the managed
   closure on first use.
--- a/apps/daemon/scripts/package-daemon.mts
+++ b/apps/daemon/scripts/package-daemon.mts
@@ -13,8 +13,9 @@
  * `node --import ./dist/instrument.js dist/index.js` with nothing else on disk.
  *
  * Unlike the desktop bundle (Electron `utilityProcess`, native modules rebuilt to Electron's ABI),
- * this targets plain Node: better-sqlite3 keeps the prebuild-install binary for the build host's
- * Node/OS/arch. It is therefore a same-platform artifact — build it on (or for) each target.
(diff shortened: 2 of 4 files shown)


PeronGH (approved):
Approving — this is necessary, not cosmetic. Without it the next tagged release fails at `verify-artifacts`.

Verified against a fresh clone of this branch with a real staging closure (`pnpm --filter @linkcode/desktop --prod deploy --legacy --cpu=x64`, pnpm 11.10.0):

- The closure has exactly one better-sqlite3, and `prebuilds/` holds all eight flat files — `darwin-arm64.node`, `win32-x64.node`, … The `prebuilds/${nodePlatform}-${arch}.node` template matches exactly (note it is *not* the prebuildify directory convention).
- **There is no `build/` directory at all**, so the old `build/Release/better_sqlite3.node` assertion would fail on every artifact.
- It won't come back via `@electron/rebuild`: v13's `binding.gyp` gates both targets on `force_build==1 or prebuild_exists==0`, where `prebuild_exists` shells out to `node lib/binding.js`. In the staged package that prints `1`, so both targets collapse to `type: none` and a plain `node-gyp rebuild` produces nothing.
- The prebuild loads with no build step — `getPrebuildPath()` → `prebuilds/linux-x64.node`, and `new Database(':memory:')` + `pragma('journal_mode = WAL')` returns SQLite 3.53.4.

The revised `deps/**` rationale in `electron-builder.yml` also checks out: `binding.gyp` `includes: ['deps/common.gypi']` unconditionally, so gyp needs `deps/` to parse even in the no-op case.

Two non-blocking points.

### 1. The check is now close to tautological

Nothing prunes non-target prebuilds. `pruneStaging` (`package-app.mts:118`) only drops `.map`/`.ts`/`.md`, and the only better-sqlite3 exclusion in `electron-builder.yml` is `deps/**`. So all eight prebuilds ship in **every** artifact and all eight smart-unpack.

That means `prebuilds/darwin-arm64.node` is present in the Windows build too, and its header is arm64 by construction — `readBinaryArch` can never disagree with `expectedArch`. Under v12 the check was genuinely target-sensitive; it now only proves the tarball wasn't mangled.

Fix that restores it *and* saves space: prune the seven non-target prebuilds in `pruneStaging`. That's ~14.3 MB of dead weight per artifact (16.5 MB shipped, 2.2 MB used), and once only the target prebuild survives, the existence + arch assertion means something again.

### 2. Comment describes a failure mode that doesn't exist yet

`verify-artifacts.mts:87` — "what breaks is the staging prune keeping the wrong target" — nothing currently prunes prebuilds per target. Accurate only *after* the change above.

PR #497, opened 2026-08-31
fix(codex): route account providers explicitly

lucas77778:
## Summary

- select a thread-scoped Codex model provider for every account-backed custom endpoint instead of relying on the built-in OpenAI provider
- route Responses over HTTPS with the account token and configured base URL, with WebSocket and OpenAI login auth disabled
- keep CLI-backed 401s on the ChatGPT login flow while reporting account-backed 401s as provider authentication failures
- document and regression-test the provider routing and auth distinction

[CODE-644](https://linear.app/arcbox/issue/CODE-644/codex-ignores-custom-provider-endpoints-and-routes-all-non-official)

## Verification

- `pnpm check:ci`
- `pnpm test` — 3040 passed, 1 skipped
- targeted Codex credential, routing, and auth tests — 55 passed
- real Codex app-server probe observed `POST /v1/responses` with the expected bearer token; a provider 401 remained recoverable and did not emit `authentication_failed`

## Checklist

- [x] If a wire message changed: `WIRE_PROTOCOL_VERSION` is bumped (no wire message changed)


xiaoland (approved):
Verified, it's work:

Image

and the code LGTM.

PR #498, opened 2026-08-31
fix(mobile): surface ambiguous deletion transport

xiaoland:
## Summary

- return an explicit unknown outcome when DELETE /account produces no HTTP response
- show neutral copy without claiming success, failure, or safe retry
- keep local deletion teardown limited to confirmed accepted responses

Follow-up to #481 for CODE-292 D-38.

## Verification

- pnpm check:ci
- pnpm vitest run --project mobile apps/mobile/src/runtime/cloud/__tests__/deletion.test.ts
- pnpm test (3038 passed, 1 skipped, before rebasing the identical patch onto current master)


lucas77778 (approved):
LGTM

PR #500, opened 2026-09-01
fix(ui): make command palette instant

lucas77778:
## Summary

- remove open, close, and list-resize motion from the Command+K palette while retaining backdrop dimming
- let the open palette handle Command+K itself so the shortcut toggles it closed; preserve Escape dismissal
- remove the workbench's now-unused Motion dependency

[CODE-648](https://linear.app/arcbox/issue/CODE-648/fixui-make-the-commandk-palette-instant-and-toggle-close)

## Verification

- `pnpm check:ci`
- `pnpm test` — 3043 passed, 1 skipped
- command-palette component tests — backdrop/no-animation, Escape close, repeated Command+K close
- browser acceptance — zero popup/backdrop animations; Command+K toggle-close and Escape close observed

## Checklist

- [x] If a wire message changed: `WIRE_PROTOCOL_VERSION` is bumped

### DIFF
--- a/packages/client/workbench/package.json
+++ b/packages/client/workbench/package.json
@@ -27,7 +27,6 @@
     "foxact": "^0.3.10",
     "foxts": "^5.9.1",
     "lucide-react": "catalog:",
-    "motion": "^12.42.2",
     "pathe": "^2.0.3",
     "posthog-js": "^1.406.2",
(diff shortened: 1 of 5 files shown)


xiaoland (approved):
LGTM.

Image

